PR #5, opened 2017-12-03
Use actor/eventemitter model on client & add channel UI

heyitsmeuralex:
image

image

See code for explanation of actor system. It's essentially a bunch of EventEmitters that talk to eachother.

PR #10, opened 2017-12-03
Add API to fetch old messages in a channel

towerofnix:
Please note: this makes message dates be stored as numbers (the output of `Date.now`) rather than ISO-formatted strings. @heyitsmeuralex, you will probably have to update #5. (Just remove `Date.parse` - see this PR's changes in `site/js/index.js`.)

### DIFF
--- a/api.js
+++ b/api.js
@@ -36,10 +36,6 @@ module.exports = function attachAPI(app, {io, db}) {
     return user
   }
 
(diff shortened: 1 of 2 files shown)


towerofnix:
Fixed that in your PR myself so you don't need to! @heyitsmeuralex

PR #32, opened 2017-12-03
Add username-available API

towerofnix:
For the username creation dialog, which will constantly fetch `/api/username-available/(usernameInput.value)` while typing and show a string "available" or "not available".


heyitsmeuralex:
>constantly fetch

oh i was just going to do it at the validation stage

PR #36, opened 2017-12-04
Remove openpgp library

### LINKED ISSUE #35: Remove openpgp library
At the moment the library is being loaded but not used.

~~I'll keep the API backends for now, since they'll be useful for future reference.~~

towerofnix:
Resolves #35.

This library wasn't being used. ~~The API methods related to it are still kept, for future reference.~~ Git works for that!


heyitsmeuralex:
>for future reference

We're using git - you can always look back at the history. Why keep these methods?

towerofnix:
Fair enough. I'll remove those.

towerofnix:
I'm going to *assume* removing that code is not cause for needing re-approval :)

PR #42, opened 2017-12-04
Fix text sometimes being displayed as black

### LINKED ISSUE #39: Make default color on dark elements be something that's not solid black
I'm particularly thinking of the message container element, whose `color` is not set. Not that there should ever be a child whose color isn't set, but - still!

(Super minor nitpick.)

towerofnix:
Fixes #39.

There's not currently a particular case where any text is affected by this, but this gets rid of the potential for accidentally black-on-black text.

### DIFF
--- a/site/styles.css
+++ b/site/styles.css
@@ -26,6 +26,7 @@ body, html, #app {
 
   font-family: sans-serif;
 
+  color: var(--text-white);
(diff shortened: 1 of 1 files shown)


heyitsmeuralex:
Wtf there are so many errors in your code please fix

Issue #17, opened 2017-12-03
Pinned messages


towerofnix:
Just like Discord - but without the 50 pin limit.

- [x] Pin messages API
- [ ] Unpin messages API
- [x] List pin messages API

towerofnix:
*Jumping* to a pinned message, which is part of Discord, can't be implemented until #31 is done.

PR #44, opened 2017-12-04
Add pinned message API

towerofnix:
Towards #17.

Send a POST request to `/api/pin-message` including your session ID and the message you want to pin to add a message to its channel's pinned message list. Only administrators can pin messages.

Send a GET request to `/api/channel/:channelID` to get the list of pinned messages (which is not returned through `/api/channel-list`):

Screenshot of outputted JSON

This doesn't do anything on the client-side.

### DIFF
--- a/api.js
+++ b/api.js
@@ -78,9 +78,13 @@ module.exports = function attachAPI(app, {io, db}) {
     // Extra details for a channel - these aren't returned in the channel list API,
     // but are when a specific channel is fetched.
     channelDetail: async c => Object.assign(serialize.channelShort(c), {
-      pinnedMessages: await Promise.all(c.pinnedMessageIDs.map(
-        async id => serialize.message(await db.messages.findOne({_id: id}))
-      ))
+      // Null messages are filtered out, just in case there's a broken message ID in the
+      // pinned message list (e.g. because a message was deleted).
+      pinnedMessages: (
+        (await Promise.all(c.pinnedMessageIDs.map(id => db.messages.findOne({_id: id}))))
+          .filter(Boolean)
+          .map(serialize.message)
+      )
     })
   }
 
--- a/api.js
+++ b/api.js
@@ -48,44 +48,48 @@ module.exports = function attachAPI(app, {io, db}) {
   }
 
   const serialize = {
-    message: m => ({
+    message: async m => ({
       id: m._id,
       authorUsername: m.authorUsername,
       authorID: m.authorID,
       date: m.date,
       channelID: m.channelID,
-      revisions: m.revisions.map(serialize.messageRevision),
+      revisions: await Promise.all(m.revisions.map(serialize.messageRevision)),
       reactions: m.reactions
     }),
 
-    messageRevision: r => ({
+    messageRevision: async r => ({
       text: r.text,
       signature: r.signature,
       date: r.date
     }),
 
-    user: u => ({
+    user: async u => ({
       id: u._id,
       username: u.username,
       permissionLevel: u.permissionLevel,
     }),
 
-    channelShort: c => ({
+    channelShort: async c => ({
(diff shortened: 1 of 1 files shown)


api.js
```diff
@@ -70,9 +70,15 @@ module.exports = function attachAPI(app, {io, db}) {
...
     }),
 
-    channel: c => ({
+    channelShort: c => ({
       id: c._id,
       name: c.name
+    }),
+
+    // Extra details for a channel - these aren't returned in the channel list API,
+    // but are when a specific channel is fetched.
```
heyitsmeuralex: Good thinking 🎉

api.js
```diff
@@ -306,6 +306,42 @@ module.exports = function attachAPI(app, {io, db}) {
...
+        response.status(500).end(JSON.stringify({
+          error: 'this message is already pinned'
+        }))
+
+        return
+      }
+
+      await db.channels.update({_id: channel._id}, {
+        $push: {
+          pinnedMessageIDs: messageID
```
heyitsmeuralex: May be worth checking to see if the message ID is actually a valid message ID and is in that channel!

api.js
```diff
@@ -77,8 +77,10 @@ module.exports = function attachAPI(app, {io, db}) {
 
     // Extra details for a channel - these aren't returned in the channel list API,
     // but are when a specific channel is fetched.
-    channelDetail: c => Object.assign(serialize.channelShort(c), {
-      pinnedMessageIDs: c.pinnedMessageIDs
+    channelDetail: async c => Object.assign(serialize.channelShort(c), {
```
heyitsmeuralex: > async

Ouch!

Would be nice for _all_ `serialize` methods to be async for consistency.

api.js
heyitsmeuralex: To be fair only administrators can pin messages, but I still think it's worth checking beforehand so we don't get any edge cases in the future if we forget to check there.

api.js
```diff
@@ -77,8 +77,10 @@ module.exports = function attachAPI(app, {io, db}) {
...
+      pinnedMessages: await Promise.all(c.pinnedMessageIDs.map(
+        async id => serialize.message(await db.messages.findOne({_id: id}))
```
heyitsmeuralex: ...what happens if the database method resolves with `null`?

towerofnix: Good news for people who love bad news :tada:

Should it just filter out null messages? - The only case I can think of where a message is null is if it's been deleted. That's not in the API yet, but it might be in the future (though I'm not too certain how to go about that). Still good to be careful.

api.js
towerofnix: Actually, both of those things are already done.

A message is always pinned in the channel it's contained in. Notice how there's no `channelID` parameter? That's because the channel is automatically fetched from the passed message:

```js

+    (req, res, next) => {
+      const v = req[middleware.vars]
+      v.channelID = v.message.channelID
+      next()
+    },
+    ...middleware.getChannelFromID('channelID', 'channel'),
+ 
```

The message is always valid, because this line (gracefully, i.e. 404) fails if it isn't:

```js
 +    ...middleware.getMessageFromID('messageID', 'message'), 
```

api.js
towerofnix: Agreed.

api.js
heyitsmeuralex: Yeah, I think its worth checking for deleted messages at this point. Deleted as in `null` 📦

api.js
heyitsmeuralex: Oman I didn't even notice that -- _awesome_!

api.js
towerofnix: Fixed!

api.js
towerofnix: Fixed.

Issue #25, opened 2017-12-03
Separate client from its host server


heyitsmeuralex:
The client should be able to connect to multiple servers at once. Decenteralized = good! See: [riot.im](https://riot.im/)

Needs UI @PullJosh and also big implementation, separating SessionActors per-server, etc.

Blocks #20.

PullJosh:
Currently working on some slight redesign work. Would it be useful to add non-functioning UI for this in preparation for the backend updates?

towerofnix:
@PullJosh Actually, I'm not sure there would *be* any backend updates, besides making the program which hosts the `site` folder be separate - I think the only needed change would be in the client, making the socket connect to a different host. That said, @heyitsmeuralex would know how to actually do that better than me - this will mean managing multiple sockets at once.

But all that's irrelevant to what you need to know, which I think is, yes, making a non-functioning UI would be helpful! :package:

PullJosh:
Sweet. Got to go for now, but I'll be back later today with fun design updates.

towerofnix:
Blocked by #41.

towerofnix:
Fixed by #45.

Issue #6, opened 2017-12-03
Use service workers


heyitsmeuralex:
https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API

Let's make `bantisocial` run _offline_! Service workers are the way we do that.
They also allow us to trigger notifications and basically pretend we're an actual mobile app.

PullJosh:
We can also go full-featured [PWA](https://developers.google.com/web/progressive-web-apps/) eventually, right?

towerofnix:
The serviceworkers branch has a bunch of work on this (particularly commit c5907ae8fd7a46b7173ec81aca318ac47b9e51a0), but I doubt any of that would merge well with master, plus the client is going to be rewritten (which means we'll need to rewrite the client code on that branch). I'll relabel this appropriately!

PR #45, opened 2017-12-04
Decentralize client & stop using socket.io

### LINKED ISSUE #41: Use WebSockets instead of socket.io, and fix CORS
Blocks #25! Clients currently cannot connect to remote servers not on the same domain because of CORS being disabled.

socket.io seems pretty terrible at sending the correct CORS headers, so it's probably worth transitioning to **ws** on the server and _WebSocket_ in the browser. We don't need the fancy polyfills and fallbacks socket.io implements anyway.

heyitsmeuralex:
Fixes #41 _and_ #25! 🎉

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -37,7 +37,6 @@ pre {
   </main>
 </div>
 
-<script src="/socket.io/socket.io.js"></script>
 <script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0-alpha2/katex.min.js" integrity="sha384-OMvkZ24ANLwviZR2lVq8ujbE/bUO8IR1FdBrKLQBI14Gq5Xp/lksIccGkmKL8m+h" crossorigin="anonymous"></script>
 
 <script nomodule src="script-nomodule.js"></script> <!-- for browsers not supporting es6 modules -->
--- a/site/index.html
+++ b/site/index.html
@@ -14,10 +14,19 @@ pre {
 
 <div id="app">
   <aside id="sidebar">
-    <div class="sidebar-heading">
-      <b>Channels</b>
-      <button class="btn" id="create-channel">New</button>
-    </div>
+    <section class="sidebar-section sidebar-section-light">
+      <div class="sidebar-heading">
+        <div id="server-status">Connecting...</div>
+        <button class="btn btn-visible" id="switch-server-btn">Switch server</button>
+      </div>
+    </section>
+
+    <section class="sidebar-section" id="channels-sidebar-section">
+      <div class="sidebar-heading">
+        <b>Channels</b>
+        <button class="btn" id="create-channel">New</button>
+      </div>
+    </section>
   </aside>
 
   <main id="central-pane">
--- a/site/js/ChannelsActor.js
+++ b/site/js/ChannelsActor.js
@@ -110,6 +110,7 @@ export default class ChannelsActor extends Actor {
   }
 
   bindToSocket(socket) {
+    this.socket = socket
     socket.on('created new channel', () => this.loadChannels())
   }
 
--- a/site/js/ChannelsActor.js
+++ b/site/js/ChannelsActor.js
@@ -141,7 +141,7 @@ export default class ChannelsActor extends Actor {
   }
 
   async populateSidebarList(channels) {
-    const sidebarEl = document.querySelector('#sidebar')
+    const sidebarEl = document.querySelector('#channels-sidebar-section')
 
     // Remove old channels list
     const oldListEl = sidebarEl.querySelector('#channels-list')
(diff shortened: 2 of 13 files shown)


site/index.html
```diff
@@ -14,10 +14,19 @@
 
 <div id="app">
   <aside id="sidebar">
-    <div class="sidebar-heading">
-      <b>Channels</b>
-      <button class="btn" id="create-channel">New</button>
-    </div>
+    <section class="sidebar-section sidebar-section-light">
+      <div class="sidebar-heading">
+        <div id="server-status">Connecting...</div>
```
towerofnix: Should this be "not connected" in the HTML, then set to "Connecting..." by JS?

site/js/ChannelsActor.js
```diff
@@ -118,14 +125,14 @@ export default class ChannelsActor extends Actor {
     return this.channels.find(c => c.name === channelName) || null
   }
 
-  viewChannel(channelID) {
+  async viewChannel(channelID) {
     this.activeChannelID = channelID
-    this.socket.emit('view channel', channelID)
+    //this.socket.emit('view channel', channelID) // TODO use api for this
```
towerofnix: >use api for this

Keep in mind that channel ID is stored on the server per-socket, not per-user. It actually looks like you've already written the code which handles `view channel`, so I'm wondering why this is commented out.

site/js/SessionActor.js
```diff
@@ -42,22 +61,32 @@ export default class SessionActor extends Actor {
...
   }
 
   isCurrentUser(userID) {
     if (!this.loggedIn) return false
     else return this.sessionObj.user.id === userID
   }
 
+  async switchServer(url) {
+    this.currentServerURL = url
+    this.sessionID = this.sessionIDs[url] || ''
```
towerofnix: Good thinking!

site/js/Socket.js
```diff
@@ -0,0 +1,100 @@
...
+    if (this.isDead) {
+      return
+    }
+
+    this.isDead = true
+    console.error('socket:: is dead')
+
+    // TODO:
+    // Display some kind of onscreen notification that
+    // the connection to the server has been lost.
```
towerofnix: May want to create an issue for this.

site/js/Socket.js
```diff
@@ -0,0 +1,100 @@
...
+        return
+      }
+
+      this.reconnect().then(() => {
+        // Connection re-established! :tada:
+        // (reconnect() automatically sets this.isDead = false)
+        console.info('socket:: connection re-established after death')
+      })
+    }, 2000)
+  }
```
towerofnix: This whole automatic reconnect setup is super handy, thanks.

site/js/api.js
```diff
@@ -1,5 +1,13 @@
...
+// These functions assume that serverURL is running
+// on standard HTTP - this will likely change in the
+// future (e.g. Service Workers only work on https).
+//
+// Note that WebSockets use their own protocols:
+//   http  <--> ws
+//   https <--> wss
+
+export function post(path, dataObj, serverURL) {
+  return fetch('http://' + serverURL + '/api/' + path, {
```
towerofnix: Could you just fetch `'//' + serverURL`? That way the protocol is automagically picked based on whatever protocol you're using to connect to the client. I guess that wouldn't make `ws://`/`wss://` be automatically picked, though.

site/js/index.js
```diff
@@ -17,20 +18,33 @@ const main = async function() {
...
-  // Actors get references to other actors, plus a
-  // reference to the WebSocket connection.
+  // Actors get references to other actors.
   for (const [ name, actor ] of Object.entries(actors)) {
     actor.name = name
     actor.actors = actors
-    actor.socket = socket
 
-    actor.init() // Actors should subscribe to events here.
+    actor.init() // Actors should subscribe to events from eachother here.
```
towerofnix: Typo? :package:

site/index.html
heyitsmeuralex: I guess. The result is the same, however.

site/js/ChannelsActor.js
heyitsmeuralex: Yeah, I uncommented this later afaik...?

heyitsmeuralex: Oh crap maybe not. It should be uncommented as I reimplemented it!

site/js/Socket.js
heyitsmeuralex: It took ages to do, and loads of different ways to do it.

Originally the client pingponged the server (as the server does to the client), but that proved kinda buggy.

site/js/api.js
heyitsmeuralex: We should enforce HTTPS soon anyway - it's needed for #6.

site/js/index.js
heyitsmeuralex: Typo?

towerofnix: "Eachother" should be two words - "each other".

heyitsmeuralex: o

heyitsmeuralex: @towerofnix can you fix this 📦

PR #47, opened 2017-12-04
Fix bug where sessionIDs would be an array

heyitsmeuralex:
(Includes #45!)

...and thus not actually store any data >:(


towerofnix:
Could this just be merged with #45? (I.e., merge `fix-localstorage-bug` into `decentralized`, then close this PR.)

heyitsmeuralex:
Uhhhh... yeah.

PR #48, opened 2017-12-05
Loads of fixes & minor style changes

heyitsmeuralex:
~~(Includes #45, rebase issues so this is probably a good call for a squash commit or something?)~~

See individual commits past e840c462492b95c11c9a7e935258b1f772031d3d for details.


site/js/bind-keys.js
```diff
@@ -1,25 +0,0 @@
...
-  })
-
-  document.addEventListener('keyup', evt => {
-    const keyIndex = keysRequired.indexOf(evt.keyCode)
-
-    if (keyIndex > keysDownCount) {
-      keysDownCount = keyIndex - 1
-    }
-  })
-}
```
towerofnix: RIP :coffin:

heyitsmeuralex: Yes

site/styles.css
```diff
@@ -174,14 +204,29 @@ body, html, #app {
...
+    display: inline-flex;
+    flex-direction: row;
+
     margin: 0; }
 
+    #form-container #form #chat-input {
+      resize: none;
+      flex-grow: 1; }
+
   #login-status {
```
towerofnix: Since the login-status element was removed, this rule should be.

PR #49, opened 2017-12-05
Format mentions only if the user actually exists

heyitsmeuralex:

### DIFF
--- a/site/js/MessagesActor.js
+++ b/site/js/MessagesActor.js
@@ -301,7 +301,7 @@ export default class MessagesActor extends Actor {
 
   // Formats some message text. Returns a <span>
   // element ready to be displayed.
-  formatMessageText(text) {
+  async formatMessageText(text) {
     const el = document.createElement('span')
     const { user } = this.actors.session.sessionObj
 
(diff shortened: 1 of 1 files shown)


site/js/MessagesActor.js
```diff
@@ -491,42 +491,42 @@ export default class MessagesActor extends Actor {
...
 
-      else if (char === '$' && charNext === '$' && currentToken !== 'latex') { startToken('latex'); c++; continue }
-      else if (char === '$' && charNext === '$' && currentToken === 'latex') { startToken('text'); c++; continue }
+      else if (char === '$' && charNext === '$' && currentToken !== 'latex') { await startToken('latex'); c++; continue }
+      else if (char === '$' && charNext === '$' && currentToken === 'latex') { await startToken('text'); c++; continue }
 
-      else if (char === '`' && currentToken !== 'code') { startToken('code'); continue }
-      else if (char === '`' && currentToken === 'code') { startToken('text'); continue }
+      else if (char === '`' && currentToken !== 'code') { await startToken('code'); continue }
+      else if (char === '`' && currentToken === 'code') { await startToken('text'); continue }
```
towerofnix: JS should just remove `await` and add a new `no-await` keyword, and also make all functions async by defualt :^)

heyitsmeuralex: yes

PR #51, opened 2017-12-05
Fix channelref bug

heyitsmeuralex:
Visiting `+server.com#the-default-aka-first-channel` would load messages there twice.


towerofnix:
I'm not sure why "make mentions prettier" is on this PR?

heyitsmeuralex:
>I'm not sure why "make mentions prettier" is on this PR?

I guessed it kinda related, since channelref styles are similar to mentions now. I can revert it if you like, although I don't see much point in doing that :shipit: 

image

vs

image

PR #52, opened 2017-12-05
Add server list

heyitsmeuralex:
image

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -2,23 +2,26 @@
 
 <link rel="stylesheet" href="styles.css">
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0-alpha2/katex.min.css" integrity="sha384-exe4Ak6B0EoJI0ogGxjJ8rn+RN3ftPnEQrGwX59KTCl5ybGzvHGKjhPKk/KC3abb" crossorigin="anonymous">
-<style>
-pre {
-  max-height: 100px;
-  overflow-y: scroll;
-  padding: 5px;
-  background-color: #CCC;
-  color: #111;
-}
-</style>
 
 <div id="app">
   <aside id="sidebar">
-    <section class="sidebar-section sidebar-section-light">
-      <div class="sidebar-heading">
-        <div id="server-status">Connecting...</div>
-        <button class="btn btn-visible" id="switch-server-btn">Switch server</button>
+    <section class="sidebar-section sidebar-section-light" id="server-list">
+      <div class="server server-list-heading-actual">Server list</div>
+
+      <div class="server server-selectable server-list-heading">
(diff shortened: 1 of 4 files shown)


site/index.html
```diff
@@ -2,23 +2,26 @@
...
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0-alpha2/katex.min.css" integrity="sha384-exe4Ak6B0EoJI0ogGxjJ8rn+RN3ftPnEQrGwX59KTCl5ybGzvHGKjhPKk/KC3abb" crossorigin="anonymous">
-<style>
-pre {
-  max-height: 100px;
-  overflow-y: scroll;
-  padding: 5px;
-  background-color: #CCC;
-  color: #111;
-}
-</style>
```
towerofnix: Thanks!

site/js/SessionActor.js
```diff
@@ -43,6 +46,35 @@ export default class SessionActor extends Actor {
       this.switchServer(url)
     })
 
+    let serverListOpen = false
+    document.querySelector('.server-list-heading').addEventListener('click', async evt => {
+      const serverListEl = document.getElementById('server-list')
+      serverListEl.style.setProperty('--server-list-height', ((Object.keys(this.sessionIDs).length + 1) * 54 + 32) + 'px')
```
towerofnix: MAGIC NUMBERS AAAAAAAAAAAAAAAAAAAa.

Is this necessary? Why must this be set through JS? Couldn't you just use a collapsed/open class, and have it only be short when collapsed?

site/js/SessionActor.js
```diff
@@ -75,9 +107,44 @@ export default class SessionActor extends Actor {
...
+        continue // See #server-current
+      }
+
+      const el = document.createElement('div')
+
+      el.classList.add('server')
+      el.classList.add('server-selectable')
+      el.classList.add('server-actual-option')
+
+      el.appendChild(document.createTextNode(url)) // TODO: get server name and display that instead
```
towerofnix: See #53.

site/js/SessionActor.js
heyitsmeuralex: >Magic numbers

no u, these are the heights of different elements as set in style.css. I guess they could do with some clarification...?

> Is this necessary? Why must this be set through JS?

Because I don't want to do the `max-height: auto; height: 99999px;` animation hack.

towerofnix: >animation hack

Oh, this is for animation; that makes more sense.

This could do with some explanation, yeah. After that this is fine to merge!

heyitsmeuralex: ...there is animation. Right now :shipit:

heyitsmeuralex: Done. Please re-review 📦

PR #53, opened 2017-12-05
Implement server settings and properties

towerofnix:
..as well as an API for server settings and an interface on the command line for server properties.

`GET /api/server-settings`
`POST /api/server-settings` `{sessionID: (must be an admin), patch: {name: "Blah" ...}}`
`> get|show|get-property|show-property <key>`
`> set|set-property <key> <value>`
`> list|list-properties`

`POST /api/server-settings` and `set-property` don't let you create new settings. Default server settings and server properties, as well as documentation for each setting, are stored in the new file `default-settings.js`. This file also contains the relevant code for loading the settings database; if a field is missing from the database, it will be added.

(This PR also moves the code for the "license" prompt command right below "help", since they're very similar.)


default-settings.js
```diff
@@ -0,0 +1,53 @@
...
+  })
+
+  // Server properties - these settings don't do anything until the server is
+  // restarted. These aren't publicly visible, and can only be edited through the
+  // command line prompt.
+  await setupDefaultSettings(serverPropertiesID, {
+    // HTTPS enforcement - whether or not to secure anything and everything sent
+    // between the server and client (including web sockets). This is not actually
+    // implemented yet.
+    https: 'off'
```
heyitsmeuralex: Could you provide some way to GET this value via the client? Not all properties, obviously, but `https`.

towerofnix: Sure: `GET /api/should-use-secure`

towerofnix:
Okay, pushed a (big) change. Definitely re-review!

settings.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+module.exports.setSetting = async function(settingsDB, categoryID, key, value) {
+  const settingSpec = defaultSettings[categoryID][key]
+
+  if (!settingSpec) {
+    return 'invalid key'
+  }
+
+  if (settingSpec.possibleValues) {
```
heyitsmeuralex: Will reiterate:
Would be nice to have a way to define a custom validation function. This is cool and good, but I can see some options that would require some fancy non-enum validation, such a valid integer from 1 - 100. Provided we don't want to have to make an array `[ 1...100 ]` :taco:

heyitsmeuralex: Implemented.

api.js
```diff
@@ -288,6 +292,44 @@ module.exports = function attachAPI(app, {wss, db}) {
...
+    async (request, response) => {
+      const { patch } = request[middleware.vars]
+
+      const serverSettings = await db.settings.findOne({_id: serverSettingsID})
+
+      const results = {}
+
+      for (const [ key, value ] of Object.entries(patch)) {
+        results[key] = await setSetting(db.settings, serverSettingsID, key, value)
+      }
```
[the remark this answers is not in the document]
towerofnix: Nope - they can't. `setSetting` returns an error message if you attempt to add a new key. Same deal with the console (you can't add new keys there, either).

PR #59, opened 2017-12-05
Fix #57

### LINKED ISSUE #57: Prompt modals should have their inputs disabled while validating
When registering on @heyitsmeuralex's testing server, I typed out the name `florrie`. I pressed enter, to select that as my username. Nothing appeared to happen for about half a second, and I'd already started typing the password I wanted to use. By the time the prompt finished verifying and the password box opened, I'd already entered two of the characters in my password - `ab`, for example. So it wasn't obvious to me if, once I confirmed the register (i.e. entered a password), my username would be `florrieab` or `florrie`. So I canceled the register. This confusion seems unnecessary. (Also, if it would be `florrieab`, and florrieab *was* taken, I wouldn't know, since the verification checked `florrie`!)

While verifying, in general, inputs should not be changeable. When an input is being verified, a snapshot of the value is taken, which is passed to the server; so, if the user changes that value on their end *after* the verification request is sent, the server will respond based on the value originally passed to it, which could lead to errors such as the given username not being available. And, of course, if the user expects the next prompt to instantly appear, they'll start typing; that's what leads to the input's value being changed after pressing enter in the first place.

Related to #34 - this should probably be done first/separately, though, so that there's still a two-part prompt to work on.

heyitsmeuralex:
See #57 - prompt modals should disable their inputs while validating input.
Also adds styling for disabled `.input` elements.

### DIFF
--- a/site/js/SessionActor.js
+++ b/site/js/SessionActor.js
@@ -256,7 +256,7 @@ export default class SessionActor extends Actor {
           const { available } = await get('username-available/' + name, this.currentServerURL)
 
           if (!available) {
-            throw 'Username unavailable.'
+            throw 'Username not available: ' + name
           }
         },
         'Continue', 'Cancel')
--- a/site/js/ModalsActor.js
+++ b/site/js/ModalsActor.js
@@ -144,14 +144,13 @@ export default class ModalsActor extends Actor {
           inputEl.disabled = true
 
           try {
-            await new Promise(resolve => setTimeout(resolve, 2000))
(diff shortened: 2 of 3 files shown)


site/js/SessionActor.js
```diff
@@ -256,7 +256,7 @@ export default class SessionActor extends Actor {
           const { available } = await get('username-available/' + name, this.currentServerURL)
 
           if (!available) {
-            throw 'Username unavailable.'
+            throw 'Username not available: ' + name
```
towerofnix: :+1:

PR #64, opened 2017-12-05
Add online offline status

### LINKED ISSUE #63: Online/offline status in backend
..based on whether any socket of the user has recently pinged.

(ping @heyitsmeuralex.. no pun intended :package:)

towerofnix:
Includes #62! Merge that one before this.

Fixes #63.

This adds a simple online/offline status system. Any serialized user will include whether the user is online or not (in the `online` field). In order to do this, I added a `sessionID` field to socket data; this is filled in by the client's `pong data` event, which is sent in response to the server's `ping for data` event. (`ping for data` is sent as soon as the socket connects, and on the ping interval.) In order to see if a user is online or not, the API simply checks for any socket whose session ID belongs to that user.

### DIFF
--- a/api.js
+++ b/api.js
@@ -671,6 +671,15 @@ module.exports = async function attachAPI(app, {wss, db}) {
     }
   ])
 
+  app.get('/api/user-list', async (request, response) => {
+    const users = await db.users.find({})
+
+    response.status(200).end(JSON.stringify({
+      success: true,
+      users: await Promise.all(users.map(serialize.user))
+    }))
+  })
+
   app.get('/api/username-available/:username', [
     ...middleware.loadVarFromParams('username'),
 
--- a/api.js
+++ b/api.js
@@ -46,6 +46,17 @@ module.exports = async function attachAPI(app, {wss, db}) {
     return user
   }
 
+  const isUserOnline = async function(userID) {
+    // Simple logic: a user is online iff there is at least one socket whose
+    // session belongs to that user.
+
+    const sessions = await db.sessions.find({userID})
+
+    return Array.from(connectedSocketsMap.values())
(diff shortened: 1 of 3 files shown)


towerofnix:
On second look this doesn't actually *do* anything with the commit in #62. Guess I shoudln't have branched from that, hm?

api.js
```diff
@@ -46,6 +46,17 @@ module.exports = async function attachAPI(app, {wss, db}) {
...
+  const isUserOnline = async function(userID) {
+    // Simple logic: a user is online iff there is at least one socket whose
+    // session belongs to that user.
+
+    const sessions = await db.sessions.find({userID})
+
+    return Array.from(connectedSocketsMap.values())
+      .some(socketData => sessions
+        .some(session => session._id === socketData.sessionID))
+  }
```
heyitsmeuralex: not webscale enough tbh

api.js
```diff
@@ -770,25 +783,46 @@ module.exports = async function attachAPI(app, {wss, db}) {
           return
         }
 
-        connectedSocketsMap.set(socket, {
+        Object.assign(connectedSocketsMap.get(socket), {
           channelID: data // channelID
         })
+      } else if (evt === 'pong data') {
+        // Not the built-in pong; this event is used for gathering
+        // socket-specific data.
```
heyitsmeuralex: Hey, uh, `socket.ping` can take a data argument. Check out the ws documentation.

towerofnix: I didn't see a way to detect the actual ping event from the browser, though? https://stackoverflow.com/a/45929287/4633828

heyitsmeuralex: Ok!

PR #72, opened 2017-12-10
Fix code blocks using multiple <code> elements

### LINKED ISSUE #71: Mark code blocks as white-space: pre
At the moment, they aren't, which makes this..

```
for (let i = 0; i < 100; i++) {
  alert(
    'haha pranked' + i
  )
}
```

..display as this:

```
for (let i = 0; i < 100; i++) {
alert(
'haha pranked' + i
)
}
```

heyitsmeuralex:
Also fixes #71

### DIFF
--- a/site/styles.css
+++ b/site/styles.css
@@ -391,6 +391,9 @@ body, html, #app {
       font-weight: bold;
       color: var(--text-red); }
 
+    .message-inline-code {
+      white-space: pre; }
+
 a {
   color: #E080FF;
   text-decoration: none; }
--- a/site/js/MessagesActor.js
+++ b/site/js/MessagesActor.js
@@ -526,7 +526,7 @@ export default class MessagesActor extends Actor {
       else if (char === '`' && currentToken !== 'code') { await startToken('code'); continue }
       else if (char === '`' && currentToken === 'code') { await startToken('text'); continue }
 
-      else if (char === '\n') {
+      else if (char === '\n' && currentToken !== 'code') {
         const state = currentToken
         await startToken('newline')
         await startToken(state)


site/styles.css
```diff
@@ -391,6 +391,9 @@ body, html, #app {
       font-weight: bold;
       color: var(--text-red); }
 
+    .message-inline-code {
+      white-space: pre; }
+
```
towerofnix: So what I'm getting is that we support `\n` in `` `abc` ``-style code blocks, and that ```` ``` ```` code blocks just aren't a thing. I wonder if this should be changed - using two line breaks in a `` `code` `` element `isn't

supported` in Markdown.

In any case, this class name definitely dosen't make sense!

```
blah:

`
asdf
baz
batr
`
```

^ This definitely isn't inline code, is it :)

heyitsmeuralex: >So what I'm getting is that we support \n in `abc`-style code blocks,

Unintentional, but yes. And there's not really much point in fixing it in my opinion.

> ``` code blocks just aren't a thing

They are-- try them.

>  this class name definitely dosen't make sense!

Will fix

towerofnix: >> triple-backslash code blocks just aren't a thing
>
> They are-- try them.

Hm? I don't see any code written to handle ```` ``` ````. Sure, using that string of characters *works*, but I don't think *that's* intentional - rather it just ignores the first two ticks and treats the third as the beginning of the "inline" code-block.

heyitsmeuralex: 100% intentional 😜 

I _thought_ I implemented it at some point but clearly not.

heyitsmeuralex:
I'll probably end up rewriting formatting. It was a really quick job before and it has some pretty big flaws.

It doesn't tokenize or anything which is a bit bad 🌮

towerofnix:
>It doesn't tokenize or anything which is a bit bad :taco:

\>`startToken`

wow is you're code lying :anger:

PR #86, opened 2017-12-12
Respond with channel ID in create-channel

### LINKED ISSUE #85: Make create-channel API respond with channel ID
..to be consistent with the `send-message` API.

I don't think the response of `create-channel` is being used by the official client's code anywhere.

towerofnix:
Fixes #85.

Before, the channel wasn't even being serialized before being responded - yikes!

(Branch title gore.. :shipit:)

### DIFF
--- a/api.js
+++ b/api.js
@@ -561,7 +561,7 @@ module.exports = async function attachAPI(app, {wss, db}) {
 
       response.status(201).end(JSON.stringify({
         success: true,
-        channel
+        channelID: channel._id
       }))
     }
   ])


api.js
```diff
@@ -561,7 +561,7 @@ module.exports = async function attachAPI(app, {wss, db}) {
 
       response.status(201).end(JSON.stringify({
         success: true,
-        channel
+        channelID: channel._id
```
heyitsmeuralex: oof

PR #91, opened 2017-12-13
Add badges to README 

vutondesign:
Adds some badges to the README

capture


heyitsmeuralex:
@PullJosh @towerofnix What do you think about this? I'm not sure if we either want badges or would like more, or what.

towerofnix:
@vutondesign This project isn't on the Awesome list, though?

PullJosh:
While I appreciate the help from @vutondesign, I think my preference would be for a badge-free readme.

towerofnix:
@PullJosh :+1:. If we were using a test/automatic build system, this would be handy, but we're not using either of those (yet). Thanks for the contribution though, @vutondesign :)

You might want to checkout some of the issues labelled ["warrants discussion"](https://github.com/towerofnix/bantisocial/issues?q=is%3Aissue+is%3Aopen+label%3A%22warrants+discussion%22) - verbal input is definitely handy. And of course, if you have any ideas, making issues for them is welcome. (And you should do that before you write the code - that way we can be sure on how we want to implement the feature, and no time is wasted on writing code that'll need to be changed later.)

vutondesign:
👍

Issue #66, opened 2017-12-05
Channel manager


towerofnix:
Related to #14.

There needs to be a way to rename and delete channels (and probably reorder them) for admins.

## Todo

- [ ] Rename channels.
  - [ ] Backend. Probably: `POST /api/rename-channel {channelID, newName, sessionID}`.
  - [ ] Frontend. Either a literal "Rename" button, or maybe an edit (pencil) symbol beside the name.
- [ ] Delete channels?
  - This is complicated. We currently go out of our way to *not* delete messages - past revisions of edited messages are even stored and viewable. Following that, it doesn't make sense to delete an entire history of messages. Should we go with an "archive" option, instead? Archived channels are hidden from the obvious channel list, but they are not removed; anybody may view an archived channel, probably through a separate list.
- [ ] Reorder channels?
  - I'm for doing it The Discord Way, which means a global order that's unsortable to the client. But shrug. See discussion in #14.

PullJosh:
> Either a literal "Rename" button, or maybe an edit (pencil) symbol beside the name.

Given that we have so many actions that we want users to be able to do (rename, delete, reorder, etc), I propose that each channel name has a three-dot context menu on the right side (only visible to admins). Something like this:

![image](https://user-images.githubusercontent.com/5458180/33629472-25f45f6e-d9d2-11e7-9daa-21b2417861e4.png)

(Clicking it, of course, would open a menu with a list of your possible actions.)

PullJosh:
Re: Deleting channels

Personally, I think it makes the most sense to provide both "archive" and "delete" options. Archived channels could still be linked to and viewed, as well as re-activited by the admin if they want. However, they would be read-only and wouldn't appear in the regular channel list. This option would be best in the vast majority of cases.

However, I'm also in favor of giving as much power as possible to the users (particularly the trusted ones, like admins). Much like in the decision **not** to artificially limit the number of custom emoji that can be uploaded, I see no reason to prevent an admin from doing something (ie deleting a channel) if they think it's the best course of action. Admins are, presumably, trusted users, so there's no reason to believe that our own judgment as developers is any better than theirs. It's quite possible that some extenuating circumstances could arise that would make permanently deleting a channel the best possible option, so there's no reason to artificially prevent it.

heyitsmeuralex:
>  I'm also in favor of giving as much power as possible to the users (particularly the trusted ones, like admins)

This is a good general design policy to go with - e.g. per-user channel order. :+1:

[...]

PR #96, opened 2017-12-14
Emit socket event on channel rename

### LINKED ISSUE #90: Send event to client sockets when channel renamed or deleted
Towards #66.

This will be needed so that the message sidebar can be automatically updated with the new titles (or with deleted channels removed). There's already a `created new channel` event.

towerofnix:
Fixes #90. Towards #66.

This emits a socket event to all clients `renamed channel { channelID, newName }` when a channel is renamed.

There's also a small reference implementation for the client code to handle this event.

### DIFF
--- a/api.js
+++ b/api.js
@@ -668,6 +668,10 @@ module.exports = async function attachAPI(app, {wss, db}) {
 
       await db.channels.update({_id: channelID}, {$set: {name}})
 
+      sendToAllSockets('renamed channel', {
+        channelID, newName: name
+      })
+
       response.status(200).end(JSON.stringify({
         success: true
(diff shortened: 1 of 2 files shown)


api.js
```diff
@@ -640,6 +640,10 @@ module.exports = async function attachAPI(app, {wss, db}) {
 
       await db.channels.update({_id: channelID}, {$set: {name}})
 
+      sendToAllSockets('renamed channel', {
+        channelID, newName: name
+      })
```
heyitsmeuralex: wow that was easy

towerofnix: haha yes it was

PR #98, opened 2017-12-14
Delete channel API

### LINKED ISSUE #97: Delete channels (backend)
Towards #66.

Channels can only be deleted by admin users. When a channel is deleted, so are all of the messages whose channel ID matches it. A socket event should be emitted when a channel is deleted.

towerofnix:
Fixes #97.

Adds a new channel for deleting APIs: `POST /api/delete-channel {channelID, sessionID}`

And a corresponding socket event: `deleted channel {channelID}`

Adds a quick change to the client to reload the channel list when a channel is deleted, though it should be noted that it, if the deleted channel was selected, a different channel won't automatically be selected.

### DIFF
--- a/api.js
+++ b/api.js
@@ -646,6 +646,33 @@ module.exports = async function attachAPI(app, {wss, db}) {
     }
   ])
 
+  app.post('/api/delete-channel', [
+    ...middleware.loadVarFromBody('channelID'),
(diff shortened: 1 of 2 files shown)


heyitsmeuralex:
>Adds a new channel for deleting APIs

🤔

Issue #18, opened 2017-12-03
Custom emoticons


towerofnix:
Defined on the server. This will require uploading images - ping @eq since you've done a bunch of image uploading service work before! :tada:

~~No 50-emote limit, but maybe there should still be *some* limit -- does 500 seem reasonable? :)~~

heyitsmeuralex:
How about a default emoji set of some kind..?

i.e. if something sends literally the `thinking` emoji unicode character what happens to it? The problem is many browser+os combinations don't really handle emoji that well:

![image](https://user-images.githubusercontent.com/9429556/33526416-d1fb178a-d838-11e7-9d2b-88595e90ee81.png)

I don't think forcing a download of a bunch of emoji is particularly worth it, but we need to discuss this.

heyitsmeuralex:
I think it'd be nice to _not_ have a default set but allow admins to define unicode aliases for emoticons they upload themselves.

i.e. if I upload a `:blobthinking:` emoji, I should be able to assign it to the literal :thinking: unicode character, so I see thinking as :blobthinking: images.

towerofnix:
>I think it'd be nice to *not* have a default set but allow admins to define unicode aliases for emoticons they upload themselves.

That's a pretty cool idea! I do think there should still be a way to add separate emoticons, though - I don't think there's a unicode emoji which would match `:gnu_slash_thinking:` very well. (Besides :thinking:, I suppose, but what if I want `:gnu_slash_thinking:` *and* `:blobthinking:`?)

heyitsmeuralex:
Oh, of course. You don't _have_ to define the alias!

heyitsmeuralex:
So what's the census here?

Default set or no?

heyitsmeuralex:
>does 500 seem reasonable? :)

surely the owner of the server -- the admins!! -- should be able to decide that

towerofnix:
My opinion is to have a default set - so that people discover the feature - but have it be really, really easy to remove the emojis from that.

>surely the owner of the server -- the admins!! -- should be able to decide that

Sure; default to 500, have a toggle to let there be no limit, which only the server owner can change. (Hum, that means adding a new permission level. But having an owner level makes sense anyways.)

heyitsmeuralex:
>owner level

Why?

Owner has access to the CLI - that's ownership enough.

towerofnix:
Okay fine, yes.

[...]

PR #100, opened 2017-12-14
Add image upload API

### LINKED ISSUE #99: Static image folder for uploads
There should be some means of uploading image files to the server. I'm thinking `/images/<md5>/<name of image>`. There should also probably be some file size limit - 10MB sounds fine? (Note this is mostly just for custom emotes and maybe profile pictures.)

Blocks #18.

towerofnix:
Issue/PR 100 :tada:

Fixes #99. Towards #18.

`POST upload-file?sessionID=.. <FormData {image: file}>`. Responds with `{success: true, path}` if successful, where `path` is the string you'd enter into your location bar (or an ``) to load the image.

Uses [Multer](https://github.com/expressjs/multer) to upload files. Files are saved according to the format `__dirname/uploads/<shortid>/<basename of original name>.<extension based on MIME type>`. Files can only be uploaded by admins (as this is mainly for custom emotes), and cannot be greater than 10MB in size (in case an admin accidentally selects the wrong file; obviously small emote icons should never be this big). Only MIME types `image/png`, `image/gif`, and `image/jpeg` are uploaded. (By the way: this means *animated* custom emotes! NYAN)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,8 @@
 db/*
 !db/.gitkeep
+
+uploads/*
+!uploads/.gitkeep
+
 node_modules
 site/script-nomodule.js
--- a/api.js
+++ b/api.js
@@ -12,8 +12,15 @@
 
 const express = require('express')
 const bodyParser = require('body-parser')
+const multer = require('multer')
+const shortid = require('shortid')
 const uuidv4 = require('uuid/v4')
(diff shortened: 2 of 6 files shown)


towerofnix:
deal with it . thisandgain

heyitsmeuralex:
svg tho

PR #106, opened 2017-12-16
Update Riot client to work with new API

towerofnix:
I can't believe you were developing this with the API *that* old :tada:


heyitsmeuralex:
xd

Issue #110, opened 2017-12-16
Rewrite client with oof


towerofnix:
Hopefully this won't be as big of an ordeal as rewriting it in Riot (#77) - I suppose it depends on how implementing things with [oof](https://github.com/heyitsmeuralex/oof) will be.

I'm thinking we could just add oof as a submodule, underneath the `site` folder? Otherwise we have to figure out a jillion other things like bower and stuff which all won't work with the whole "no building" thing (which is basically the point of oof).

@heyitsmeuralex Obviously we need to design oof, too. I'm thinking avoid dealing with styling - don't bother with scoped CSS (and whatever scary things it might imply, such as shadow DOM and browser incompatibilities). That way we can just reuse ~~[the old CSS](https://github.com/towerofnix/bantisocial/blob/551f15fae247022ba5ec4a012ae37697c94318eb/site/styles.css)~~ [PullJosh's CSS](https://codepen.io/PullJosh/pen/OOYZWK). No need to rewrite *everything!*

PullJosh:
> That way we can just reuse [the old CSS](https://github.com/towerofnix/bantisocial/blob/551f15fae247022ba5ec4a012ae37697c94318eb/site/styles.css). No need to rewrite everything!

You're going to have a lot more success reusing the [current CSS](https://codepen.io/PullJosh/pen/OOYZWK). :P

TheInitializer:
@PullJosh wow, that looks really nice!

PR #111, opened 2017-12-17
Delete .travis.yml

towerofnix:
It's spamming our emails, and we aren't actually going to use it in the near future (see #110).


heyitsmeuralex:
Setup gh-pages to run on master, then. We can direct people to towerofnix.github.io/bantisocial/site.

Issue #101, opened 2017-12-14
Profile pictures


towerofnix:
The first thing that comes to (my) mind when you say "profile pictures" is *hosting*. Then again, does that matter? Collaterally, messages weigh a *lot* more than some 100x100 images.

Alternatively, [Libravatar](https://www.libravatar.org/) could provide a solution for having profile pictures without hosting them on the chat server. But that's not necessarily good, because it means depending on an external service, which some clients might not be capable of handling (e.g. a restricted network connection where Libravatar is blocked/not whitelisted).

So I'm biased more towards hosting ourselves, but this maybe warrants discussion.

~~Partially blocked by #77.~~

I'm assigning..

* Myself, for backend work. If we go with Libravatar, there'll need to be work to integrate it and "connect" Libravatar accounts to users. If we host the files ourselves, we'll need to have an API to update and fetch profile pictures.

* @PullJosh, to design how profile pictures should look on the client. I think you might have done some work on the codepen sketch already?

* @heyitsmeuralex, to actually implement that and put together the API and design.

heyitsmeuralex:
>to design how profile pictures should look on the client.

PullJosh already did this, and it's implemented in the `riot` branch.

heyitsmeuralex:
I would say we should use Libravatar as many sites use it (or Gravatar, which it fallbacks to).

>If we go with Libravatar, there'll need to be work to integrate it and "connect" Libravatar accounts to users.

The only thing we need is the email address of each user. Libravatar works by taking an md5 hash of their email address.

towerofnix:
> ? Do you mean implement it?

Sorry, I meant for you to write the frontend stuff :package: - which it looks like you've already done.

heyitsmeuralex:
> verifying emails

Also, that's pointless 😉 
I say we don't store users' email addresses and just md5 hash them when their email is set.

Issue #68, opened 2017-12-06
User list UI


heyitsmeuralex:
Would be nice to have a list of users and their online/offline states on the right side of the screen a la every other chat application in existence!

###### Once again @PullJosh if you write some HTML/CSS for this I'll make it actually function.

heyitsmeuralex:
This has been done, now we just need to implement it. See #77

Issue #14, opened 2017-12-03
Channels should have order


heyitsmeuralex:
At the moment they're just ordered by wherever they are in the database.

Question: **should channel order be**:
* global? (a la Discord)
* per-user? (a la Riot)

heyitsmeuralex:
Also there needs to be a client-side way to change channel order if we choose either global or per-user.

towerofnix:
Why global? Perhaps because this way everybody has the same order - it's consistent. This means that admins can organize things in a way that Makes Sense.

Why per-user? Users get to choose what way Makes Sense - to *them*. Going per-user would mean the order isn't consistent, though. (Also, there would need to be a default order - probably alphabetical.)

Why permanent? Nobody's opinions would have a play, and the order would be objectively predictable, if you can recall the names of the channels. It would also certainly be the easiest option to implement. However, permanently sorting by a particular method means no form of categorization (either global or per-user).

heyitsmeuralex:
>Also, there would need to be a default order - probably alphabetical.

We can consider combining global and per-user. That is, the admins set a default but users may change it if they like.

towerofnix:
Perhaps - I don't want to make the UI too complicated, though. A "restore to server default" button would work, but it would also be nice to preview that. Or, we could have two separate "tabs" - one for the global layout, one for per-user (which could be copied from default at any point).

heyitsmeuralex:
>we could have two separate "tabs"

:-1:

I think we need to add some kind of settings page that lets you change the order so we don't overcomplicate/saturate the UI. Under the admin area of the settings admins would be able to update the default order for the whole server.

PR #112, opened 2017-12-17
Oof

### LINKED ISSUE #110: Rewrite client with oof

towerofnix:
Fixes #110 (or at least begins it).

Demo screenshot

Yeow! This is one big change. Don't feel obligated to review it super indepth-ly.

This gets rid of Riot and Travis (so it closes #111), gets us a *lot* closer to [PullJosh's design](https://codepen.io/PullJosh/pen/OOYZWK), and simplifies the client's code a bunch. It's both smaller in volume (only 300 lines, not including websocket.js and oof, and only one JavaScript file) and in complexity (the code is relatively much simpler than working with Riot, and somewhat even simpler than the old actor model (#5)). Also, error stacks work; there's no compilation voodoo of *any* kind going on here.

While this has more features than the Riot client, it's not a complete implementation of the client (before either of those). Still to be done - but in separate PRs (and in no particular order):

- *Message* `formatting!`
- Save and load data from local storage.
- Edit and delete messages.
- Handle channel rename/delete events.
- Improve date formatting. Showing the full timestamp isn't particularly friendly.
- Prompt/error modals.
- Implement the user list. (This is tracked in #68.)
- Implement avatars. (This is tracked in #101).
- Fix the order channels show up in. (This is tracked in #14.)

NYAN


heyitsmeuralex:
General points:
* cool and good
* Could do with some more comments
  - especially for 'sections' of the code, e.g. `////////// CHANNEL LIST //////////`

site/index.js
```diff
@@ -0,0 +1,352 @@
+'use strict'
+
+// Utility functions:
+
+function post(path, dataObj) {
+  if (!serverURL.value) {
+    return {error: 'client error - server URL not specified yet'}
```
heyitsmeuralex: Surely this should be `return Promise.reject({...})`?

site/index.js
```diff
@@ -0,0 +1,352 @@
...
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify(dataObj)
+  }).then(res => res.json())
+}
+
+function get(path, query = {}) {
+  if (!serverURL.value) {
+    return {error: 'client error - server URL not specified yet'}
```
heyitsmeuralex: Same here

site/index.js
```diff
@@ -0,0 +1,352 @@
...
+        appendMessage(data.message)
+      }
+
+      if (evt === 'created new channel' && data && data.channel) {
+        sidebarChannelList.append(data.channel)
+      }
+    }
+  })
+
+  serverDict.socket = socket
```
heyitsmeuralex: This should be `serverDict[serverHostname].set('socket', socket)` or something. We want to have a separate socket connection kept open _for each server at the same time_.

site/index.js
```diff
@@ -0,0 +1,352 @@
...
+
+  activeServerHostname.set(serverHostname)
+}
+
+document.getElementById('add-server').addEventListener('click', () => {
+  const host = prompt('Host URL?')
```
heyitsmeuralex: oof no modals

site/index.js
towerofnix: Probably. It should be noted that server errors also return data ala `{error: 'asdf'}`. Maybe we could make `fetchHelper` automatically throw responses which include an `error` field?

site/index.js
towerofnix: :+1:, that was a typo!

site/index.js
towerofnix: Nope, not yet.

site/index.js
towerofnix: Fixed. Errors are now a lot more error-y. This also simplified the code a fair amount!

towerofnix:
>Could do with some more comments, especially for 'sections' of the code

Good point! Thanks. This should've been obvious but 01:00 me didn't think of it. I've reorganized the code, putting similar things closer to each other, and everything in neat commented sections.

PR #117, opened 2017-12-17
Update README to include --recursive in git clone

towerofnix:
Otherwise, `oof` won't be cloned. Git submodules aren't unflawed :package:


PullJosh:
> aren't unflawed

:thinking:

Issue #114, opened 2017-12-17
Message formatting


towerofnix:
Furthering #110.

This was originally added in #1 for the pre-Riot client. Now that we've gone through about a billion rewrites, we need to implement message formatting again.

heyitsmeuralex:
[mrk](https://github.com/heyitsmeuralex/mrk) has been made for this purpose. Extensions we need:

* [x] Strikethrough
* [ ] Images
* [ ] Channel/server references
* [ ] @mentions

heyitsmeuralex:
Images have been done in #130.

heyitsmeuralex:
Going to defer mentions to #92.

heyitsmeuralex:
![image](https://user-images.githubusercontent.com/9429556/34297315-60be3134-e70f-11e7-8f3a-30a84be1af80.png)

PR #126, opened 2017-12-19
Choo 🚂🚋

### LINKED ISSUE #31: Implement channel scrollback
Scrolling to the top of a channel's message list should load more messages from that channel.

I've already made an API for this: `GET /api/channel/:channelID/latest-messages/before/:messageID`. Message elements from that list should be built the way the elements of the newest messages are, and should be appended above (before) the oldest message element.

When should scrollback message loading occur? It would be easy to make it happen as soon as the screen is scrolled all the way to the top, but maybe fetching and appending messages a little earlier would be better (so as to make scrolling smoother).

When the scrollbar is *dragged* (rather than the scrollwheel being used), should messages only be loaded once the user releases the scrollbar? This way the scrollbar will stay static until it's not immediately being used anymore (so it's easier to navigate to the message you're looking for).

### LINKED ISSUE #34: Combine username and password dialogs
..into one dialog, which requests both your username and password in the same view.

This applies for both login and register.

### LINKED ISSUE #113: Load session data and channel list from localStorage
Furthering #110. These should be saved and loaded from local storage.

### LINKED ISSUE #116: Improve message date formatting
#112 implemented a basic timestamp system, showing every message group's date. It displays the full date-time string, though, which is often more than wanted. See #15 for discussion.

### LINKED ISSUE #120: Use the latest version of oof
[oof has been updated](https://github.com/heyitsmeuralex/oof/wiki) a whole bunch. It's now a whole lot more like React, in some ways:

```js
class Counter extends oof.El {
  init({ startingCount = 0, changeBy = +1 }) {
    this.count = new oof.Value(startingCount)

    this.interval = setInterval(() => {
      
      // Every second, update the count
      this.count.set(this.count.value + changeBy)

    }, 1000)

    return [ this.count ]
  }

  render(count) {
    const text = document.createTextNode(`Count: ${count}`)
    const span = document.createElement('span')

    span.appendChild(text)

    // On click, reset count back to zero.
    span.addEventListener('click', () => {
      this.count.set(0)
    })

    return span
  }

  destroy() {
    // Clean up
    clearInterval(this.interval)
  }
}

const myCounter = new Counter('#my-counter', { changeBy: +1 })

myCounter.count.onChange(count => {
  console.log(count)
})
```

We definitely want:
* oof
* ~~morphdom~~
* ~~Some kind of **bel**-like library. Warrants discussion!~~ justel

### LINKED ISSUE #122: Client rewrite war!
Blocks *literally every other issue.* (Except maybe some backend ones.) Should fix a few.

Here we go...

### What on earth is the client rewrite *war*??????????

This is a *fun, competitive project* in which a bunch of people (well, @PullJosh and @heyitsmeuralex) write clients for Decent. The "best" one is merged into this repository and used as the official one.

### What kinds of clients?

We're looking to make a replacement for the current "official" client program - that's the one that exists within the `towerofnix/decent` repository. This means that the client just one basic condition: **The client should work in your web browser.** Mind you, using recent ES2017/whatever features is okay; just as long as your client works inside a (normal: firefox, chrome, not IE) browser.

### Can I use Browserify?

Yes. Use *Babel*, if you want to - I don't care.

### What does my client need to do?

There's only one *essential* guideline, which I stated above: the client should work in your web browser. You can't replace the existing client with [a plugin for weechat](https://github.com/towerofnix/decent/issues/79), sorry. :)

However, there *is* a list of things your client must be able to do. Remember that *you are writing a replacement,* so your code needs to do everything that the current client does *and some more* (so that there's bonus incentive to use your client, rather than what is currently here).

Furthermore, **your client must be based on @PullJosh's [design](https://codepen.io/PullJosh/pen/OOYZWK).**

Good news for you, [the current client](https://github.com/towerofnix/decent/tree/34288140b32eb10f70f95876fadb4877b3807552/site) is very simple. However, we want more than what's there. Here's a checklist for the features you should support (~~copy/paste it into your PRor whatever - don't click the buttons here~~ progress on alex's PR #126):

- [x] Support selecting a server. The server dropdown should be filled.
- [x] Support viewing channels. The sidebar channel list should automatically be filled when you select a server, and should automatically update according to the [appropriate](https://github.com/towerofnix/decent/wiki/API#from-server-created-new-channel) [websocket](https://github.com/towerofnix/decent/wiki/API#from-server-renamed-channel) [events](https://github.com/towerofnix/decent/wiki/API#from-server-deleted-channel).
- [x] Support registering. **There should be some sort of a modal - do *not* use `prompt`.** Handle error messages appropriately (unfortunately there's no documentation for them, but you can read [the code](https://github.com/towerofnix/decent/blob/34288140b32eb10f70f95876fadb4877b3807552/api.js#L895-L934); it's pretty self descriptive and easy to follow). You should confirm that the entered username [is available](https://github.com/towerofnix/decent/wiki/API#get-apiusername-availableusername) before actually sending [the register request](https://github.com/towerofnix/decent/wiki/API#post-apiregister).
- [x] Support logging in and out. Logging in means making [the appropriate request](https://github.com/towerofnix/decent/wiki/API#post-apilogin) and keeping track of the returned session ID; logging out means discarding that session ID. These should update the "user info" display (username, log in/out/register buttons), obviously.
  - [x] Support handling the [`ping for data`](https://github.com/towerofnix/decent/wiki/API#to-client-ping-for-data) WebSocket event. You can confirm this is working by checking [the user list endpoint](https://github.com/towerofnix/decent/wiki/API#get-apiuser-list): when you're logged in, the user of the session ID you're logged in with should show as "online" there.
- [x] Support sending messages. Pressing enter in the text field should [submit a message](https://github.com/towerofnix/decent/wiki/API#post-apisend-message) (and so should clicking the "send" button).
- Support reading messages.
  - [x] Read messages as they're sent by handling [the relevant WebSocket event](https://github.com/towerofnix/decent/wiki/API#from-server-received-new-message).
  - [x] Fetch the most recent messages. [There's an API for that.](https://github.com/towerofnix/decent/wiki/API#get-apichannelchannelidlatest-messages)
  - [ ] Fetch old messages (the same API, but with `?before=(messageID)` specified) when you scroll to the top. This is *not* something you can skip. See #31 for info.
  - Messages are displayed in groups. These follow PullJosh's design. (You can use placeholder profile pictures for now.) Group timestamps should be formatted according to https://github.com/towerofnix/decent/issues/116#issuecomment-352268366. To determine if a message should appear in the same message group as another, the following conditions should be true:
    - The two messages should have the same author (ID).
    - The two messages should be consecutive. Message A shouldn't appear in the same group as Message C if Message B is already in a different group (e.g. because it has a different author).
    - The message group should never contain more than 20 messages.
    - If the later message was posted at least an hour after the earlier one, it shouldn't be in the same group.
- [x] Store and load session data (and the added server list) from `localStorage`. See #113. When you sign in, you should stay signed in until you click the "log out" button, even if you reload the page.

I'm picking all of the above things for two reasons, generally. The first is that it'll get things done - several of those have never been implemented before. The second, more important one, is that it'll show the strengths of using one framework - really, one client's code - over the other. These are fairly complicated and varied tasks (generate message groups and load them on scrollback, handle logged in/out state, interact with the API).

Have fun and *please* do ask me about anything related in comments/on matrix (ideally here; if on matrix, ping me).

heyitsmeuralex:
Resolves #122 if merged :dash:

* Closes #120 
* Furthers #114 
* Fixes #113 
* Fixes #116 
* Fixes #34 
* Fixes #31


site/js/util/modal.js
```diff
@@ -0,0 +1,131 @@
...
+    // construct #modal element
+    this.el = html`<div class='modal'>
+      <div class='modal-close-button' onclick=${this.close.bind(this)}></div>
+      <div class='modal-header'>${opts.title} <span class="modal-header-subtitle">${opts.subtitle || ''}</span></div>
+      ${content}
+    </div>`
+
+    // darken page cover
+    const pageCover = document.querySelector('.modal-page-cover')
+    pageCover.onclick = this.close.bind(this) // close on click
```
towerofnix: Code style nit - this (and other cases of `this.[foo].bind(this)`) should be `() => this.close()`.

site/js/util/modal.js
```diff
@@ -0,0 +1,131 @@
...
+    pageCover.classList.add('visible')
+
+    // show the modal
+    this.visible = true
+    document.body.insertBefore(this.el, pageCover)
```
towerofnix: The modal element goes *before* the page cover?

site/js/util/modal.js
```diff
@@ -0,0 +1,131 @@
...
+
+  focus() {
+    // focus on the FIRST input
+    // tabindex is set so people can then use tab to select the next one
+    this.styledInputs[0].el.querySelector('input').focus()
+  }
+
+  get values() {
+    // map of input id to its current value
+    const map = {} // we _would_ use a Map here but then you can't destructure it
```
towerofnix: Yes, you can: `for (const { name, value } of this.values.entries())`.

site/js/util/modal.js
```diff
@@ -0,0 +1,131 @@
...
+    // this only means emitting the 'submit' event
+    this.emit('submit', this.values)
+  }
+
+  close() {
+    // close the modal
+    this.visible = false
+    this.el.remove()
+
+    // darken page cover
```
towerofnix: Typo? More like "un-darken page" or "hide page cover".

site/js/util/storage.js
```diff
@@ -0,0 +1,20 @@
...
+
+  set(key, value) {
+    const valueJSON = JSON.stringify(value)
+
+    localStorage.setItem(key, valueJSON)
+    console.log('stored', { key, value })
+  },
+}
+
+module.exports = storage
```
towerofnix: Cool stuff! :tada:

site/js/util/api.js
```diff
@@ -0,0 +1,39 @@
+// wrapper around window.fetch
+
+async function fetchHelper(host, path, fetchConfig = {}) {
```
towerofnix: This should throw an error if `host` is not given. (Actually, it should also throw an error if path isn't given, but a missing host argument is more likely - the path is usually a static string given straight into the function call, while the host is usually gotten from some variable, which might have an undefined/null value.)

site/js/content/index.js
```diff
@@ -69,12 +69,12 @@ content.use((state, emitter) => {
...
-
-    // listen for events
-    state.ws.on('received chat message', handleNewMessage)
-    state.ws.on('edited chat message', handleEditMessage)
+    if (host) {
+      // establish/use websocket
+      state.ws = new Ws(host)
+      state.ws.on('received chat message', handleNewMessage)
+      state.ws.on('edited chat message', handleEditMessage)
+    }
```
towerofnix: This was causing an error when you navigated to the page without initially giving a server path (i.e. `localhost:3000`, not `localhost:3000/my.cool.server:12345`) so I fixed it.

site/js/util/modal.js
heyitsmeuralex: Yep.

site/js/sidebar/channel.js
```diff
@@ -0,0 +1,85 @@
...
+      try {
+        const { success } = await api.post(host, 'create-channel', {
+          name: channelName.trim(),
+          sessionID: state.sessionID,
+        })
+
+        modal.close()
+      } catch (error) {
+        if (error.message === 'name invalid') {
+          modal.showError('Invalid name - can only contain a-z, 0-9,\nunderscores, and dashes')
```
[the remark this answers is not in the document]
heyitsmeuralex: yep. blame pulljosh

site/js/util/modal.js
towerofnix: Why?

index.js
```diff
@@ -41,6 +41,12 @@ async function main() {
   await setupDefaultSettings(db.settings)
   await attachAPI(app, {wss, db})
 
+  // Any other route points to index.html.
+  // The client uses pushState-based routing, so this is required!
+  app.get('*', (request, response) => {
+    response.sendFile(__dirname + '/site/index.html')
+  })
+
```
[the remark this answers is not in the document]
heyitsmeuralex: gh-pages? no

ehh.

site/js/index.js
```diff
@@ -0,0 +1,14 @@
+// import utils
```
towerofnix: For what it's worth, it would be nice if `api` were imported and published (i.e. `window.api === require('./util/api')`), since it makes debugging/testing things quite a bit easier :)

site/js/content/index.js
```diff
@@ -0,0 +1,201 @@
...
+      const wasScrolledToBottom = (messagesContainer.scrollTop > getScrollDist() - 50)
+
+      state.messages.push(message)
+      emitter.emit('render')
+
+      // autoscroll after we render if we were scrolled to the bottom beforehand
+      if (wasScrolledToBottom)
+      setTimeout(() => {
+        messagesContainer.scrollTop = getScrollDist()
+      }, 25)
```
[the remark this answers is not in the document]
heyitsmeuralex: >there's not apparently an event for that

I might make a PR for it.

site/js/content/message-editor.js
```diff
@@ -0,0 +1,75 @@
+const html = require('choo/html')
+const api = require('../util/api')
+
+module.exports = (state, emit) => {
+  const textarea = html`<textarea
+    class='message-editor-input'
+    placeholder='Enter a message...'
+    onkeyup=${keyup}>
+  </textarea>`
```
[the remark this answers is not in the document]
heyitsmeuralex: It's actually bel.

site/js/content/message-editor.js
```diff
@@ -0,0 +1,75 @@
...
+    // shift
+    if (key === 16) {
+      state.shift = true
+    }
+
+    // enter
+    if (key === 13) {
+      if (state.shift) {
+        // if shift is down, enter a newline
+        // this is default behaviour
```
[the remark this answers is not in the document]
heyitsmeuralex: .............oh yeah!

site/js/util/modal.js
heyitsmeuralex: The use-case for modal.value is in `on('submit', ({ username, password })` - you can't nicely destructure a Map there.

site/js/content/message-editor.js
```diff
@@ -0,0 +1,75 @@
...
+    const editor = html`<div class='message-editor'>
+      ${textarea}
+      <button class='message-editor-button' onclick=${send}>Send</button>
+    </div>`
+
+    editor.isSameNode = (a) => {
+      // classname will be different between this state & not logged in
+      // we don't really need to morph this dom element very much
+      if (editor.className === a.className) return true
+    }
```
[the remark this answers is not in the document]
heyitsmeuralex: I hastily wrote this to fix the bug we found where the textarea would get updated and cleared. Will clarify.

> and it looks like you should probably be doing this instead

...that would have the same effect? `isSameNode` is used to figure out (quickly) if the node should be morphed or not - we only ever need it to morph it when we render the other branch of this if, i.e. the className is not the same.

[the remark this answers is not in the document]
towerofnix: It's more explicit: "if we are currently logged in, set this element to only update when we are not logged in" rather than "if we are currently logged in, set this element to only update when the class name changes".

site/js/util/modal.js
towerofnix: Ah, cool!

site/js/sidebar/index.js
```diff
@@ -226,6 +225,13 @@ sidebar.use((state, emitter) => {
       history.replace(`/${host}/#${newName}`)
     }
   }
+
+  async function pongData() {
+    console.log('hey-ho', state.sessionID)
```
heyitsmeuralex: ?

index.js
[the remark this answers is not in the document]
towerofnix: Hosting on a static site *would* be nice, though, hm?

site/js/sidebar/index.js
towerofnix: See https://github.com/towerofnix/decent/pull/126/commits/bb720c6ff059d353ac1d937f35f0878d8403c51b

towerofnix: ..oh lol

site/js/index.js
towerofnix: I'll fix this now.

towerofnix: Done!

towerofnix:
All good except a couple code things I can change if you want, and one issue with scrollback: my scroll position isn't recovered such that the *previous* last message is still visible; I'm scrolled immediately to the top of the *just-added* messages. (I think that's just because your prepending messages, which naturally moves the elements I was previously scrolled over down.)

index.js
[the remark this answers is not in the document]
towerofnix: Oh, cool, you fixed this. :tada:

heyitsmeuralex:
> one issue with scrollback: my scroll position isn't recovered such that the previous last message is still visible; I'm scrolled immediately to the top of the just-added messages.

weird - I don't get that. unless I'm misunderstanding

site/js/util/modal.js
heyitsmeuralex: For some reason PullJosh's CSS only works when it's inserted before the page cover. It doesn't really matter, though.

site/js/util/modal.js
heyitsmeuralex: We can't do the following:

image

towerofnix:
Kay, fixed that bug in https://github.com/towerofnix/decent/pull/126/commits/b53b58cabed8e25b404bd5c1f9e8f966575db14a (but you should view it [without whitespace diffing](https://github.com/towerofnix/decent/pull/126/commits/b53b58cabed8e25b404bd5c1f9e8f966575db14a?w=1)).

site/js/content/message-editor.js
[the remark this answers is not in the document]
towerofnix: I'll fix this now.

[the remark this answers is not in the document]
towerofnix: Done.

towerofnix:
@heyitsmeuralex Ready for you to merge.

PR #129, opened 2017-12-21
Add account settings page/api endpoint and user profile pictures (avatars)

### LINKED ISSUE #101: Profile pictures

heyitsmeuralex:
Resolves #101 :tada:  
This PR merges into `componentize`, **not** master, to help with merge conflicts. It therefore blocks it.

Uses libravatar, and stores email addresses on the server. You get a default retro avatar based on the md5 hash of your user ID:

image

And you can use the account settings page (accessed by clicking your username in the session info area in the sidebar) at `/servers/:host/account-settings` to edit the email address associated with your account.

image

Note that "Change password" just runs `alert('not implemented')`.

Any further messages from you will take your avatar URL with them (`msg.authorAvatarURL`):

image

### API changes

User objects now have a string `authorAvatarURL` field. Likewise, message objects have `authorAvatarURL`, set at the time they are created. Note: in the database these 'avatar URLs' don't exist - they are stored as `email` and `authorEmail` (null, by default) and avatar URLs are only created when the object is seralized.

**New endpoint!** `POST /api/account-settings { email: String, sessionID: String }`. Returns `{ success: true, avatarURL: ... }`

`/api/session/:sessionID` now gives you `user.email`.

### DIFF
--- a/api.js
+++ b/api.js
@@ -19,6 +19,8 @@ const fs = require('fs')
 const path = require('path')
 const util = require('util')
 const bcrypt = require('./bcrypt-util')
+const crypto = require('crypto')
+const memoize = require('memoizee')
 
 const mkdir = util.promisify(fs.mkdir)
 
@@ -41,6 +43,17 @@ module.exports = async function attachAPI(app, {wss, db}) {
       socket.send(JSON.stringify({ evt, data }))
     }
   }
+
+  const md5 = string => {
+    if (!string) {
+      throw 'md5() was not passed ' + string
+    }
+
+    return crypto.createHash('md5').update(string).digest('hex')
+  }
+
+  const emailToAvatarURL = memoize(email =>
+    `http://cdn.libravatar.org/avatar/${email ? md5(email) : ''}?d=retro`)
 
   const getUserBySessionID = async function(sessionID) {
     const session = await db.sessions.findOne({_id: sessionID})
@@ -111,6 +124,7 @@ module.exports = async function attachAPI(app, {wss, db}) {
       id: m._id,
       authorUsername: m.authorUsername,
       authorID: m.authorID,
+      authorAvatarURL: emailToAvatarURL(m.authorEmail || m.authorID),
       text: m.text,
       date: m.date,
       editDate: m.editDate,
@@ -121,6 +135,7 @@ module.exports = async function attachAPI(app, {wss, db}) {
     user: async u => ({
       id: u._id,
       username: u.username,
+      avatarURL: emailToAvatarURL(u.email || u._id),
       permissionLevel: u.permissionLevel,
       online: await isUserOnline(u._id)
     }),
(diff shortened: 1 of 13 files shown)


api.js
```diff
@@ -42,6 +44,17 @@ module.exports = async function attachAPI(app, {wss, db}) {
...
+    if (!string) {
+      throw 'md5() was not passed ' + string
+    }
+
+    return crypto.createHash('md5').update(string).digest('hex')
+  }
+
+  const emailToAvatarURL = memoize(email =>
+    `https://seccdn.libravatar.org/avatar/${email ? md5(email) : ''}?d=retro`)
+
```
towerofnix: This will cause mixed content warnings on browsers using a client hosted on `http` (not secure), I think? Not a big deal, but it would be nice to set this based on whether or not the server's HTTPS setting is true or not.

api.js
```diff
@@ -1040,9 +1078,15 @@ module.exports = async function attachAPI(app, {wss, db}) {
         return
       }
 
+      const serializedUser = await serialize.user(user)
+
+      // We will also provide the stored email address, but only
+      // because this IS the user and it is therefore safe to give it away.
+      Object.assign(serializedUser, { email: user.email || null })
+
```
towerofnix: I'm not feeling great about attaching data outside of `serialize` - perhaps `serialize.user` could take a sessionID property? Compare with how we attach unread messages count to channels when a session ID is passed:

```js
channelBrief: async (c, sessionUser = null) => {
  const obj = {
    id: c._id,
    name: c.name
  }

  if (sessionUser) {
    obj.unreadMessageCount = await getUnreadMessageCountInChannel(sessionUser, c._id)
  }

  return obj
}
```

So we'd do basically the same with users:

```js

user: async (u, sessionUser = null) => {
  const obj = {
    id: u._id,
    username: u.username,
    avatarURL: emailToAvatarURL(u.email || u._id),
    permissionLevel: u.permissionLevel,
    online: await isUserOnline(u._id)
  }
  
  if (sessionUser && sessionUser._id === u._id) {
    obj.email = u.email || null
  }

  return obj
}
```

(That means doing `serialize.user(user, user)`, which admittedly looks a little silly, but *is* consistent with `serialize.channel`! :shipit:)

api.js
heyitsmeuralex: No, it's the other way around. But I agree.

towerofnix: Ah, okay, cool!

PR #132, opened 2017-12-21
Fix server properties not persisting

heyitsmeuralex:
image

Was caused by a missing 'async' keyword. Surprised this went unnoticed for so long!


towerofnix:
>Was caused by a missing 'async' keyword. Surprised this went unnoticed for so long!

Huh! Oops. Probably because we weren't actually using any server settings until just now.

Issue #92, opened 2017-12-13
Mentions


heyitsmeuralex:
We need to store unread @mentions on the backend so we can display unread ones to the user. Like Discord. I'd like to label this as vital as honestly it's a feature I _really_ miss using Matrix.

 It's also worth keeping in mind that the way Matrix/Riot does this is the same as IRC, which is just to gather a list of mentions **client-side** and highlight them differently. This means you can't see mentions from the history of the room you're in, or globally :(

---

Discord implements this by formatting messages specially before they are sent by the client. It will convert `@alex` to `@<1234567890>` (or whatever _alex_'s user ID is) and then send _that_ to the server. The server will parse `@<[0-9]+>`s present in the message text and add that mention to the database. It'll also ping the user.

This is cool, because it ensures that even if you change your username - which you can't at the moment! - your mentions will stay intact. Oh, and we can't forget `@everyone`!

This can go along with a "mentions only" channel notifications feature (another thing to add!!), as opposed to just 'mute' or 'all.'

towerofnix:
>We need to store unread @mentions on the backend so we can display unread ones to the user. Like Discord. I'd like to label this as non-vital as honestly it's a feature I really miss using Matrix.

Absolutely!

(Did you mean *vital* (rather than *non-*vital) though? :package:)

>Discord implements this by formatting messages specially before they are sent by the client. It will convert @alex to @<1234567890> (or whatever alex's user ID is) and then send that to the server. The server will parse @<[0-9]+>s present in the message text and add that mention to the database. It'll also ping the user.

I wonder if it would be better to process/change this on the server? It seems like a pain to force clients to handle this themselves. (Though, of course, you could also send `@<asdf>` from the client, and that would work too.)

heyitsmeuralex:
I don't think processing it on the server is a good idea, for the simple reason that only the client knows who `@alex` is referring to when there are two `alex`s! This is currently not something that could happen but I think it's more flexible to let the client do it.

PullJosh:
Discord has a place where users can view all unread mentions. Is that something we'd like to include in our UI as well? ;)

towerofnix:
@PullJosh Sure, sounds reasonable to me, though you might want to consider the "collapsed" state - will it be obvious that the right sidebar can act as anything other than the user list when collapsed? (Of course, you couldn't view your mentions in a sidebar 80 pixels wide, but - *when collapsed*, it should be clear that *when not collapsed* you can use it as a way to see old mentions.)

[...]

PR #133, opened 2017-12-22
Even more formatting

### LINKED ISSUE #114: Message formatting

heyitsmeuralex:
image

Resolves #114. Does **not** add formatting for mentions as we need backend work for those - discussion going on in #92.


site/src/util/mrk.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    read(numBackticks)
+    if (look() !== '\n' && look() !== '') return false
+
+    // Set metadata
+    meta({ lang, code })
+
+    return true
+  },
```
towerofnix: I tried to implement codeblocks the other day and failed miserably. This looks nice! :tada:

heyitsmeuralex: Was that what towerofnix/mrk was for?

towerofnix: Yup. I think it was also to make some internal changes to support my thing (which is usually a bad idea anyways :P).

PR #135, opened 2017-12-23
Emotes

### LINKED ISSUE #18: Custom emoticons

heyitsmeuralex:
image

image

Resolves #18


towerofnix:
Blocked by #133.

site/src/app.js
```diff
@@ -43,7 +43,10 @@ app.use((state, emitter) => {
     state.secure = false
     state.ws = new util.WS(state.params.host)
 
-    state.ws.on('open', () => state.secure = state.ws.secure)
+    state.ws.on('open', () => {
+      state.secure = state.ws.secure
+      emitter.emit('emotes.fetch')
```
heyitsmeuralex: Note that this should probably be auto-fetched periodically (or when a new `emotes update` event is sent over the WS) on clients. At the moment you need to reload in order to see newly-added emoji.

site/src/components/messages.js
```diff
@@ -266,6 +268,13 @@ const component = (state, emit) => {
     }
   }
 
+  if (state.messages.isScrolledToBottom()) {
+    // scroll to bottom after re-render
+    setTimeout(() => {
+      state.messages.el.scrollTop = state.messages.el.scrollHeight + 999
+    }, 50)
+  }
+
```
towerofnix: Why the `+ 999`? Shouldn't `scrollHeight` automatically adjust enough?

heyitsmeuralex: Yep, but it didn't.

JS is bad.

PR #141, opened 2017-12-24
Private servers

### LINKED ISSUE #136: Password-protected/invite-only servers

towerofnix:
Resolves #136.

Servers aren't private by default; this can be enabled by setting `requireAuthentication` to `on` through the command line (since it's a server property). It can be checked with the API endpoint `GET /api/require-authentication` (which responds with `{serverRequiresAuthentication: true/false}`).

The only way to authorize a user at the moment is to make them an admin. There will be an API endpoint and UI for this later.

---

@heyitsmeuralex I'll quote commit 8c6531e113e3460c0cf28f124561e5068c404f80 here since I think it's relevant to you:

> **Add 'routeready' event**
>
> This is emitted *after* app.js does everything it needs to do with the route event. It's super handy and fixes like a billion (not filed) bugs.
>
> Without this, we have to choose a "sensible default" for whether the server is private or not, and whether it runs on HTTPS or not. But that default isn't actually right - you end up having mixed content, or errors related to fetching an API without passing a session ID, or even just not having messages fetched at all, etc. This gets rid of those by giving a different event - routeready - for those actions to wait on.
>
> PS, routeready also waits for the WebSocket to be ready, since pretty much everything doesn't function correctly/completely without it (e.g. old messages showing up, but new messages not, because the WS isn't connected and so isn't reveing "received a message" events).


api.js
```diff
@@ -404,6 +409,80 @@ module.exports = async function attachAPI(app, {wss, db}) {
     next()
   })
 
+  // Don't let users who aren't verified (permissionLevel unauthorized)
+  // do anything with the API. TODO: Be able to disable this!
```
heyitsmeuralex: Would appreciate permissionLevel being a number rather than a string, by the way. e.g.

```
0 unathorized
1 member
2 admin
```

towerofnix: Yeah, I agree - but that's an outdated comment. I'm now using a separate `authorized` property (so, `authorized`: true/false).

settings.js
```diff
@@ -43,7 +43,13 @@ const defaultSettings = {
...
     // between the server and client (including web sockets). This is not actually
     // implemented yet.
-    https: {value: 'off', possibleValues: ['on', 'off']}
+    https: {value: 'off', possibleValues: ['on', 'off']},
+
+    // Authorization required - whether or not users will need to be authorized
+    // before they can interact with the server (or view its messages). Anyone
+    // can still register, but an admin must mark the user as authorized before
+    // they will be able to send or receive any information to/from the server.
+    requireAuthorization: {value: 'off', possibleValues: ['on', 'off']}
```
heyitsmeuralex: Would be nice to have an `authorizationMessage` that appears if you're logged in but not authorized. e.g. 'Contact hello@example.com for an account.'

site/src/components/srv-settings/authorized-users.js
```diff
@@ -0,0 +1,63 @@
...
+      <h1>Authorized users <span class='subtitle'>on ${state.params.host}</span></h1>
+
+      Loading...
+    </div>`
+  }
+
+  return html`<div class='page ${prefix}'>
+    <h1>Authorized users <span class='subtitle'>on ${state.params.host}</span></h1>
+
+    Hello!! Wow, look at all these users.
```
heyitsmeuralex: kek

api.js
```diff
@@ -1109,14 +1109,45 @@ module.exports = async function attachAPI(app, {wss, db}) {
...
+          : Promise.resolve(null)
+      ])
+
+      const result = {
+        success: true,
+        users: await Promise.all(authorizedUsers.map(serialize.user))
+      }
+
+      // Respond the unauthorized users in a separate field, but only if the
+      // session user is an admin.
```
heyitsmeuralex: :+1:

settings.js
towerofnix: Yeah, good idea!

towerofnix: johndoe

site/src/components/sidebar.js
```diff
@@ -1,6 +1,8 @@
 // sidebar component
 const html = require('choo/html')
+const raw = require('choo/html/raw')
 const css = require('sheetify')
+const mrk = require('../util/mrk')
```
heyitsmeuralex: Watch out - util/mrk supports channelrefs and stuff!

towerofnix: Not sure how to fix this without a pretty big update to mrk - e.g., temporarily disabling/enabling features for one particular `mrk` call, or setting "categories" (such as "permit `# headings` but not channel refs in the "outside of messages" category). Not *too* big of a deal, though, but still something to do later!

towerofnix:
@heyitsmeuralex This is *probably* ready for a merge, but I really do encourage you to pull and test this locally first - I might've missed a couple bugs. I wrote most of this around midnight, so, while I did write things (8c6531e113e3460c0cf28f124561e5068c404f80, 055403440545489566054b0f497e359f867a678e) to make the code a little easier to deal with, I can't guarantee I did it all perfectly!

site/src/components/sidebar.js
heyitsmeuralex: Categories can be done with `has(token, beforeToken)`

heyitsmeuralex:
Would appreciate docs for this now that they're part of the repository. Merge master into `private-server` and then make changes please 😄 

Reviewing soon <sup>tm</sup>

towerofnix:
@heyitsmeuralex Updated documentation.

site/src/app.js
```diff
@@ -55,7 +55,7 @@ app.use((state, emitter) => {
 
       state.serverRequiresAuthorization = (
```
heyitsmeuralex: ouch

site/src/components/srv-settings/authorized-users.js
```diff
@@ -185,7 +205,10 @@ const component = (state, emit) => {
 
     <p>${textarea}</p>
 
-    <p><button class='styled-button' onclick=${() => emit('authorizedUsers.saveMessage')}>Save message</button></p>
+    <p>
+      <button class='styled-button' onclick=${() => emit('authorizedUsers.saveMessage')}>Save message</button>
+      ${state.authorizedUsers.authMessageSaved ? html`<span class='status'>Saved.</span>` : ''}
```
heyitsmeuralex: :+1:

site/src/components/srv-settings/emotes.js
```diff
@@ -168,7 +168,6 @@ const component = (state, emit) => {
     </table>
 
     <div class='submit'>
-      <span class='status'></span>
```
heyitsmeuralex: is this intentional??

towerofnix: Yes: https://github.com/towerofnix/decent/pull/141/commits/b2aab1f9882ac4f2232ade657bc0b7dc2fe8b99c

That `status` element isn't used anywhere in `emotes.js`?

(Admittedly I didn't test it, so if there's an error, I wouldn't know. Oops.)

towerofnix: There was no error when I (just) tested adding an emoji.

heyitsmeuralex: Oh, I guess you're right.

I seem to remember it having a save/status feature, but clearly not - I think it's because I moved it to using modals.

heyitsmeuralex:
Error feedback would be nice.

image

heyitsmeuralex:
Seems to work thanks to those recent patch commits, thanks :shipit:
Going to actually review the code now.

api.js
```diff
@@ -259,6 +287,22 @@ module.exports = async function attachAPI(app, {wss, db}) {
       })
     ),
 
+    runIfCondition: (conditionFn, runIfSo, runIfNot = []) => (
```
heyitsmeuralex: This middleware stuff is becoming its own DSL.

api.js
```diff
@@ -404,6 +467,87 @@ module.exports = async function attachAPI(app, {wss, db}) {
...
+      // No need to rewrite the wheel - we have a fancy way of guessing where
+      // we might get the session ID, but we can use it just like we always
+      // do (with our normal middleware functions).
+      ...middleware.getSessionUserFromID('sessionID', 'sessionUser'),
+
+      (request, response, next) => {
+        // Now that we have the session user, we can actually check if the user
+        // is authorized. For now, we just check if the user's permission
+        // level is "admin" or "member", but in the future (TODO) we should have
+        // a more sophisticated (and not hard-coded!) setup for permissions.
```
heyitsmeuralex: +1

heyitsmeuralex: It may be fun to implement Discord-style granular permissions!

api.js
```diff
@@ -1088,20 +1285,97 @@ module.exports = async function attachAPI(app, {wss, db}) {
...
+
       response.status(200).end(JSON.stringify({
         success: true,
         user: serializedUser,
+
+        // This is redundant since it's already stored on the user, but it's
+        // nice to have anyways - it makes the "authorization message" property
+        // seem less out of place.
+        userAuthorized,
+        authorizationMessage
```
heyitsmeuralex: Yes, why _is_ `authorizationMessage` given at this endpoint? Surely it should be given at `should-use-authorization`?

docs/api.md
```diff
@@ -28,12 +32,16 @@ Takes the parameter `patch` and overwrites each of the specified properties acco
...
-Returns `{success: true, avatarURL}`, where `avatarURL` is a string URL (usually pointing to Libravatar) to be used as the user's profile picture.
+Returns `{success: true, avatarURL}`, where `avatarURL` is a string URL (usually pointing to [Libravatar](https://www.libravatar.org/)) to be used as the user's profile picture.
 
 ### GET `/api/should-use-secure`
 
 Returns a simple object `{useSecure}`, where `useSecure` is a boolean specifying whether or not to use WebSocket Secure and HTTPS (instead of normal WebSockets and HTTP). Of course, this assumes that HTTPS is properly set up on the host. Whether this returns true or false can be changed via the server command line (see `help` and view information on "set").
 
+### GET `/api/should-use-authorization`
+
+Returns an object `{useAuthorization}`, where `useAuthorization` is a boolean specifying whether or not to use [authorization](#authorization).
```
heyitsmeuralex: Should definitely give the authorizationMessage here.

docs/api.md
```diff
@@ -126,7 +134,23 @@ Returns `{success: true, messages}`, where messages is an array of the 50 most r
...
+
+[Authorizes](#authorization) the given user. Returns `{success: true}` if successful. Doesn't do anything (returns an error) if authorization is disabled.
+
+### POST `/api/deauthorize-user`
+
+- Parameters:
+  * `userID`: (via data; string) the unique ID of the user to be deauthorized.
+  * `sessionID`: (via data; string) the session ID of the user who is requesting this endpoint. **The requesting user must be an admin.**
+
+[Deauthorizes](#authorization) the given user. Returns `{success: true}` if successful. Doesn't do anything (returns an error) if authorization is disabled.
```
heyitsmeuralex: \+ You cannot deauthorize yourself.

Error messages in docs when?

docs/api.md
```diff
@@ -251,6 +278,29 @@ Some endpoints return further information when a session ID is given (e.g. `/api
...
+
+Authorization is a simple form of privacy which prevents clients from interacting with the server API without being authorized to do so (usually manually, by a human). This limits interaction to specific users, which may be wanted so that the server is "private". It should be noted that **enabling authorization does not encrypt messages or user data**; it simply limits who can access that data via the API.
+
+Authorization is a server property and can only be enabled via the actual command line:
+
+```
+> set requireAuthorization on|off
+```
+
+When authorization is enabled for the first time, only admins will be verified. When a user is made to be an admin through the command line (`make-admin`), they will also be authorized. (If you're using an old database, you can authorize existing admins by running `make-admin` again.) Users can then be authorized via the [`authorize-user`](#post-apiauthorize-user) endpoint (in the official client, there's a dedicated settings page for this). Users can be deauthorized using [`deauthorize-user`](#post-apideauthorize-user).
```
heyitsmeuralex: >If you're using an old database

Don't document this forever, thanks. Wait until we start doing releases before mentioning how to upgrade etc.

docs/api.md
```diff
@@ -251,6 +278,29 @@ Some endpoints return further information when a session ID is given (e.g. `/api
...
+
+When a request is made to the API of a server which requires authorization, the server searches for a session ID given in the request. (First it checks for a `sessionID` field in POST data; if that's not found, it checks the `?sessionID` query field.) If no session ID is found, or the session ID is for a user who isn't authorized, the request is immediately ended with status code 403 and an error. Otherwise, the request is processed as normal.
```
heyitsmeuralex: I feel like we should unify these into an `x-session-id` header and entirely drop these two other ways of providing the sessionID.

towerofnix:
>Error feedback would be nice.

That button is now disabled:

Greyed-out "deauthorize florrie" button

api.js
towerofnix: To be fair, yes 🎉

api.js
towerofnix: I don't think should-use-authorization existed when this was implemented - good idea!

PR #150, opened 2017-12-28
Session deletion

### LINKED ISSUE #43: Kill old sessions
When should sessions be killed, on the server? After a week of not being accessed?

In any case, sessions should most definitely be removed when the logout button is pressed.

towerofnix:
This PR contains two Big Things: automatic deletion of old sessions (fixes #43) and manual deletion of sessions. It also contains a few changes to the API and related documentation.

**Automatic deletion of old sessions** fixes #43 by pruning sessions that are at least 30 days old every five minutes and on server startup.

**Manual deletion of sessions** is a new feature which lets any user delete any sessions logged into their account. It looks like this:

List of sessions under new section "Login sessions" in account settings page, each showing the (blurred out) session ID and a "delete" button

You (and only you) can view and delete sessions from your account settings page.

Relevant API documentation can be found in commit a1396a7447fcb15e082e6ac221709309f2c9b823.

This is blocked by PR #141; I intend to update `/session/:sessionID` to return session objects, but I can't change it right now because #141 already changes the endpoint.

### DIFF
--- a/site/src/components/account-settings.css
+++ b/site/src/components/account-settings.css
@@ -85,5 +85,12 @@
   .session-id {
     color: var(--gray-300);
     font-family: monospace;
+    white-space: nowrap;
+    filter: blur(6px);
+
+    &:hover {
+      filter: blur(0);
+      transition: filter 2s;
+    }
   }
 }
--- a/site/src/components/account-settings.css
+++ b/site/src/components/account-settings.css
@@ -60,6 +60,7 @@
     margin-top: 80px;
     margin-left: auto;
     margin-right: 0;
+    margin-bottom: 20px;
     float: right;
 
     & > .status {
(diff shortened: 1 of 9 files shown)


towerofnix:
@heyitsmeuralex It would be nice if this was updated every time the `account` page was opened; I'm not sure how to do that? (It should just emit `accountSettings.fetchSessions` when `account` is opened.)

heyitsmeuralex:
>  It would be nice if this was updated every time the account page was opened

Store the `/:page` and check to see if it changed to account on `'route'` (or whatever it's called now), I think.

> but I can't change it right now because #141 already changes the endpoint.

Since we're no longer rebasing you're alright to merge that branch into this branch and continue from there. There shouldn't be any conflicts 🎉

site/src/components/account-settings.css
```diff
@@ -85,5 +85,12 @@
   .session-id {
     color: var(--gray-300);
     font-family: monospace;
+    white-space: nowrap;
+    filter: blur(6px);
```
heyitsmeuralex: cool and good

PR #156, opened 2018-01-05
Fix month formatting in date stringify util

towerofnix:
"Nov 26", not "Nov26".

### DIFF
--- a/site/src/util/date.js
+++ b/site/src/util/date.js
@@ -32,7 +32,7 @@ module.exports.timeAgo = function(date) {
 
   if (ago > week) {
     const d = new Date(date)
-    return { needsUpdate: false, string: month(d) + d.getDate() }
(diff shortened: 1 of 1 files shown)


site/src/util/date.js
```diff
@@ -32,7 +32,7 @@ module.exports.timeAgo = function(date) {
 
   if (ago > week) {
     const d = new Date(date)
-    return { needsUpdate: false, string: month(d) + d.getDate() }
```
heyitsmeuralex: whoooooooooooooooops

towerofnix: this is what happens when you don't test your code for at least a month :^)

heyitsmeuralex: did this not go wrong with, say, october?

it's just been overlooked i guess

PR #159, opened 2018-01-08
Pass state.secure directly to WS

towerofnix:
..so that WS doesn't need to call `api`. Also, this makes WS independent of Decent (so it could be used in other programs, if wanted).

This gets rid of the "Host string provided, not state object" warning.


heyitsmeuralex:
LGTM!

PR #164, opened 2018-01-08
Centralize route authentication

heyitsmeuralex:
Enables authentication with *any route* via:
* sessionID in post body
* sessionID in query string
* X-Session-ID in request headers

Before, some routes would expect sessionID to be in the post body, some in the query string, etc. This makes them not care. Documentation has been updated to match.

No frontend changes have been made, but it should be trivial to evict all `?sessionID` and `{ sessionID }` declarations in api.js calls. Just using X-Session-ID is definitely much easier.

---

@towerofnix I'd recommend some extra testing for this particular PR than usual. I don't normally work on the backend all that much, and this is an API-wide change. I may have overlooked something.

@TheInitializer this may make weecent's code simpler or something, so you might want to take a look once/if this gets merged.

### DIFF
--- a/api.js
+++ b/api.js
@@ -1359,6 +1359,8 @@ module.exports = async function attachAPI(app, {wss, db}) {
   ])
 
   app.post('/api/delete-sessions', [
+    ...middleware.loadVarFromBody('sessionIDs'),
+
     // No verification ("are you the owner of this session ID" etc), because
     // if you know the session ID, you obviously have power over it!
 
--- a/api.js
+++ b/api.js
@@ -513,32 +513,37 @@ module.exports = async function attachAPI(app, {wss, db}) {
       next()
     },
 
+    (request, response, next) => {
+      // First we check the POST body for the session ID (if it's a POST
+      // request). If we don't find anything, we check the query URL, then
+      // the headers (X-Session-ID). If we still don't find a session ID and
+      // one is required - shouldVerify - we'll say none was given
+      // and prevent the user from proceeding.
+      //
+      // See the 'authorization' section in the API docs.
+      let sessionID
+      if (request.method === 'POST' && 'sessionID' in request.body) {
+        sessionID = request.body.sessionID
+      } else if ('sessionID' in request.query) {
+        sessionID = request.query.sessionID
+      } else if ('x-session-id' in request.headers) {
+        sessionID = request.headers['x-session-id'] // All headers are lowercase.
+      } else if (request[middleware.vars].shouldVerify) {
+        // No session ID given - just quit here.
+        response.status(403).end(JSON.stringify({
+          error: 'missing sessionID field - not authorized to access API'
+        }))
+        return
+      }
+
+      // We'll save the session ID as a middleware-var so we can use it
+      // in the upcoming requests.
+      Object.assign(request[middleware.vars], {sessionID})
+
+      next()
+    },
+
     ...middleware.runIfVarExists('shouldVerify', [
-      (request, response, next) => {
-        // First we check the POST body for the session ID (if it's a POST
-        // request). If we don't find anything, we check the query URL.
-        // If we still don't find a session ID, we'll say none was given
-        // and prevent the user from proceeding.
-        let sessionID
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -1,16 +1,18 @@
...
+The project's API is publically available to anybody with access to the actual server on which it is hosted. [HTTP endpoints](#http-endpoints) provide virtually all methods of interaction from the client towards the server, while [WebSocket events](#websocket-events) let the server send messages to the client. Common data received from the server (such as users or messages) always follows [particular formats](#objects). It would be wise to understand and expect [authorization](#authorization) to be required. All endpoints, unless s General information which doesn't particularly fit anywhere else can be found in the appendix-esque section [Etc](#etc), and any questions one might have can be posted to the [issue tracker](https://github.com/towerofnix/decent/issues).
```
towerofnix: `All endpoints, unless s` Oops?

api.js
```diff
@@ -1194,11 +1184,6 @@ module.exports = async function attachAPI(app, {wss, db}) {
 
   app.get('/api/user-list', [
     ...middleware.runIfCondition(() => shouldUseAuthorization, [
-      ...middleware.loadVarFromQuery('sessionID', false),
-      ...middleware.runIfVarExists('sessionID',
-        middleware.getSessionUserFromID('sessionID', 'sessionUser')
-      ),
-
```
towerofnix: This gets rid of the `sessionUser` middleware variable so it is probably broken.

api.js
```diff
@@ -1375,8 +1359,6 @@ module.exports = async function attachAPI(app, {wss, db}) {
   ])
 
   app.post('/api/delete-sessions', [
-    ...middleware.loadVarFromBody('sessionIDs'),
-
```
towerofnix: This should not be removed - `sessionIDs`, not `sessionID`.

docs/api.md
```diff
@@ -325,6 +327,15 @@ Note that some endpoints do not require authorization:
...
 * [`GET /api/session/:sessionID`](#get-apisessionsessionid)
 
+### Authentication:
+
+When a request is made to the API, the server searches for a session ID given in the request using:
+* `sessionID` in POST data
+* `?sessionID` in query string
+* `X-Session-ID` header
+
+If the server requires [authorization](#authorization) and the session ID could not be found or pointed to an unauthenticated user, the request is immediately terminated with a 403 and error message. We recommend simply sending the `X-Session-ID` header with _all_ requests, even on non-authorization-requiring servers.
```
towerofnix: I don't like "we" - this should be `it's recommended to` 😃

api.js
heyitsmeuralex: As far as I know all authenticated endpoints get the sessionUser variable for free.

towerofnix: I mean, the only solution to this argument is to *test* `/api/user-list` (but I'm not quite awake yet). I don't see anything that's *added* to save `sessionUser`, though; only that it now saves <code>session<em>Id</em></code> for free.

docs/api.md
heyitsmeuralex: `It's probably simpler to send` thanks

docs/api.md
```diff
@@ -1,6 +1,6 @@
...
+The project's API is publically available to anybody with access to the actual server on which it is hosted. [HTTP endpoints](#http-endpoints) provide virtually all methods of interaction from the client towards the server, while [WebSocket events](#websocket-events) let the server send messages to the client. Common data received from the server (such as users or messages) always follows [particular formats](#objects). It would be wise to understand and expect [authorization](#authorization) to be required. All endpoints, unless General information which doesn't particularly fit anywhere else can be found in the appendix-esque section [Etc](#etc), and any questions one might have can be posted to the [issue tracker](https://github.com/towerofnix/decent/issues).
```
towerofnix: Still not fixed - "All endpoints, unless General information which doesn't particularly fit anywhere else can be found in the appendix-esque section..."

"General information..." is a separate sentence from "All endpoints, unless". (All endpoints unless should be removed or completed.)

heyitsmeuralex: whooops

PR #169, opened 2018-01-13
Rename websocket events; fix channel-delete handler of client

towerofnix:
Renames a bunch of WebSocket events:

* `ping for data` -> `pingdata`
* `pong data` -> `pongdata`
* `received chat message` -> `message/new`
* `edited chat message` -> `message/edit`
* `created new channel` -> `channel/new`
* `renamed channel` -> `channel/rename`
* `deleted channel` -> `channel/delete`

Updates the client to deal with these changes, as well.

Also fixes the code which handled "deleted channel"; this used to error, but now actually removes the channel from the channel list (with no error).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -206,33 +206,33 @@ These are the events which are used to send (and receive) data specific to indiv
 
 This project uses a WebSocket system which is similar to [socket.io](https://socket.io/) (though more simple). Messages sent to and from clients are JSON strings following the format `{evt, data}`, where `evt` is a name representing the meaning of the event, and `data` is an optional property specifying any additional data related to the event.
 
-### To client: `ping for data`
+### To client: `pingdata`
 
 Sent periodically (typically ever 10 seconds) by the server, as well as immediately upon the client socket connecting. Clients should respond with a `pong data` event, as described below.
 
-### From client: `pong data`
+### From client: `pongdata`
 
 Should be sent from clients in response to `ping for data`. Notifies the server of any information related to the particular socket. Passed data should include:
 
 * `sessionID`, if the client is "logged in" or keeping track of a session ID. This is used for keeping track of which users are online.
 
-### From server: `received chat message`
+### From server: `message/new`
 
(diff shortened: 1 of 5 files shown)


heyitsmeuralex:
Ping @TheInitializer - weecent will need an update.

docs/api.md
```diff
@@ -206,33 +206,33 @@ These are the events which are used to send (and receive) data specific to indiv
 
 This project uses a WebSocket system which is similar to [socket.io](https://socket.io/) (though more simple). Messages sent to and from clients are JSON strings following the format `{evt, data}`, where `evt` is a name representing the meaning of the event, and `data` is an optional property specifying any additional data related to the event.
 
-### To client: `ping for data`
+### To client: `pingdata`
 
 Sent periodically (typically ever 10 seconds) by the server, as well as immediately upon the client socket connecting. Clients should respond with a `pong data` event, as described below.
```
heyitsmeuralex:
```diff
- `pong data`
+ `pongdata`
```

docs/api.md
```diff
@@ -206,33 +206,33 @@ These are the events which are used to send (and receive) data specific to indiv
...
 
-### From client: `pong data`
+### From client: `pongdata`
 
 Should be sent from clients in response to `ping for data`. Notifies the server of any information related to the particular socket. Passed data should include:
```
heyitsmeuralex: Same here.

heyitsmeuralex:
It may be wise to have a shared module between both client and server that maps some identifier to the actual websocket event name, so we don't have naming errors in the future:

```js
const WS_EVENT = require('../ws-event')

ws.on(WS_EVENT.NEW_CHANNEL, f)
```

Thoughts? Might not be worth the hassle.

towerofnix:
>It may be wise to have a shared module between both client and server that maps some identifier to the actual websocket event name, so we don't have naming errors in the future.

Make it into a separate repository, so clients can get event names without having to include the whole server! :trollface:

Just kidding. I think it's not worth the trouble, because these names are a *lot* more sensible and consistent.

Anyhow, review requested again; I've fixed the doc links. (Try them yourself! Or don't, since I already did, and verified they work.)

PR #170, opened 2018-01-13
Fix missing padding at the bottom of .page elements

towerofnix:
This adds a margin to the bottom child element of `.page`s, so that you can scroll that distance ahead. This only really has an effect when the browser window is fairly short; before, elements like the table in the account settings page would appear right at the bottom of the screen (so the bottom of the table touched the bottom of the page).

### DIFF
--- a/site/src/app.css
+++ b/site/src/app.css
@@ -78,6 +78,14 @@ body {
       font-size: 24px;
     }
   }
+
+  /* since bottom padding doesn't actually do anything, we need to add a
+     margin onto the bottom element
+     (this only works as long as the bottom element in the page is a block-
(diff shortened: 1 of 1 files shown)


heyitsmeuralex:
We should definitely consider fetching before the actual end - Riot.im, for example, starts fetching only halfway up.

towerofnix:
Friendly ping @heyitsmeuralex; you said you would review, but you've been online and interacted with the repo since then. Review? 😃

heyitsmeuralex:
How unfriendly of me :shipit: <sup>on that note pleanse come to matrix/meta.decent.chat?</sup>

PR #171, opened 2018-01-14
Implement user online/offline internals much better

towerofnix:
This is blocked by #169 (since it includes commits from that).

This shouldn't be merged until the API docs are updated to include `user/onilne` and `user/offline`, but the PR can be reviewed now.

Copying the text from the commit:

---

This commit comes in three parts:

* Backend changes. When `pongdata` sees that a socket's session ID has changed, it announces that the previous user went offline and that the current user went online. (Note that `socketData` now keeps track of `userID`.)

* Client session object proxy. This is so that `pongdata` is sent whenever the `session` object's `id` attribute is changed (in any way).

* Client compatibility changes. Since the session object is now a proxy, it is always of the type 'object', so we need to be careful when checking if it exists (you can't do `state.session ? 'logged in' : 'logged out'` anymore).

This isn't actually exposed to the user of the client yet; this commit simply implements the internals. If you view the browser console, you'll see `ws.user/offline` and `ws.user/online` events ("no handler for ws.user/online", etc).

### DIFF
--- a/api.js
+++ b/api.js
@@ -1447,7 +1447,7 @@ module.exports = async function attachAPI(app, {wss, db}) {
 
       const { evt, data } = messageObj
 
-      if (evt === 'pong data') {
+      if (evt === 'pongdata') {
         // Not the built-in pong; this event is used for gathering
         // socket-specific data.
         if (!data) {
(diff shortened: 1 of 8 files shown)


towerofnix:
>...I assume you know what you're doing.

Me too, thanks.

PR #172, opened 2018-01-14
Transform into a monorepo

heyitsmeuralex:
See the READMEs for more information. As discussed on https://meta.decent.chat.

---

Enables the following, among other things.
```
> npm install --global @decent/cli
> decent
```

(We need to start versioning and `npm publish`ing soon!)


packages/cli/README.md
```diff
@@ -0,0 +1,24 @@
+# @decent/cli
+CLI for running a [Decent](https://github.com/decent-chat/decent) server alongside the standard web client, [@decent/client](https://github.com/decent-chat/decent/tree/master/packages/client). This is **not** a CLI client for Decent.
+
+If you want to access the server instance itself, check out [@decent/server](https://github.com/decent-chat/decent/tree/master/packages/server).
+
+### install
+```sh
+> npm install --global @decent/cli
+```
```
towerofnix: Does this actually work? I don't want to include this in the README until it's actually published to npm.

packages/cli/bin/decent.js
```diff
@@ -1,72 +1,26 @@
...
+  app.use(express.static(client)) // index.html, dist/, img/
+  app.get('*', (req, res) => res.sendFile(client + '/index.html'))
 
   const rl = readline.createInterface({
     input: process.stdin,
     output: process.stdout
   })
 
-  rl.setPrompt('> ')
+  rl.setPrompt('decent> ')
```
towerofnix: 👍

packages/client/README.md
```diff
@@ -0,0 +1,23 @@
+# @decent/client
+Standard web client for [Decent](https://github.com/decent-chat/decent).
+
+**You probably don't want to install this package directly!** If you're looking to host a [Decent](https://github.com/decent-chat/decent) chat server, you'll want [@decent/cli](https://github.com/decent-chat/decent/tree/master/packages/cli), which acts as an interface to both this package and [@decent/server](https://github.com/decent-chat/decent/tree/master/packages/server) itself.
+
+This package exports its `__dirname`. You can then use that to host `index.html`, `dist/`, `img/`, etc.
+
+If, however, you just want to self-host a client to access Decent servers with:
```
towerofnix: "with this"?

packages/client/src/util/mrk.js
```diff
@@ -82,7 +82,6 @@ const withState = state => {
       if (look() === '#') {
         read()
 
-        let c
```
towerofnix: Where is `c` defined? The `let c` from the previous block won't exist here - that `c` only exists in the scope of the block after `if (look() === '+')`. (`let` has no hoisting :) )

packages/server/api.js
```diff
@@ -28,7 +28,7 @@ const {
   serverSettingsID, serverPropertiesID, setSetting,
 } = require('./settings')
 
-module.exports = async function attachAPI(app, {wss, db}) {
+module.exports = async function attachAPI(app, {wss, db, dir}) {
```
towerofnix: `dir` here should be `dbDir`.

packages/cli/README.md
heyitsmeuralex: I'd say we publish `0.0.0` asap.

packages/client/src/util/mrk.js
heyitsmeuralex: For some reason uglifyjs complained that `c` was already defined when I did `yarn run build`.

packages/client/README.md
heyitsmeuralex: No - `If you want a client to access Decent servers with`

packages/client/src/util/mrk.js
heyitsmeuralex: Yup, I did that.

PR #187, opened 2018-01-16
Nanocomponents

### LINKED ISSUE #134: Images in messages cause auto-scrolldown to not work sometimes
### The bug
Sometimes, if an image embedded in a message takes long to load, the client doesn't automatically scroll down to the latest message when (re)loading a channel. This is probably caused by the image not taking up any vertical space until it's loaded.

It may not just be images.

### A solution

Know the height/size of each image (**how**?) before we emit render, and apply some kind of filler image whilst it is loading, but with the correct dimensions. Discord does this with a little load animation.

### LINKED ISSUE #137: Convert messages component into a nanocomponent for stability
See https://github.com/choojs/nanocomponent#nanocomponentprototypeafterupdateel.
Should only require a simple restructure and not a rewrite of business logic.

At the moment we use `setTimeout` which is a _huge_ hack!

### LINKED ISSUE #139: Don't do anything when the selected sidebar is clicked again
Clicking the selected channel at the moment causes it to be loaded, which makes the message list appear invisible for a moment (only to be replaced with the messages that were already there before). If the user really does want to reload the channel, they can select a different channel and then the other one, or reload the page.

### LINKED ISSUE #167: Change any camelCase choo event names to lowercase
I added a few camelCase events without realizing we were sticking to lowercase. Oops!

heyitsmeuralex:
**Work-in-progress.**

- Adds linting (according to [standardjs](https://standardjs.com/)); use `npm run test` to lint.
- Removed routing. It caused too much hassle
  * Now using [nanochoo](/heyitsmeuralex/nanochoo)! 🎉 
  * Yes, I know, I keep flip-flopping on this
- Server: `require('cors')` for CORS because X-Session-ID was getting eaten
- Changed the way state flows through the client
  * wow nanocomponents 🎈  <sup>not a rewrite i promise</sup>
  * Most state is now specifically per-component rather than global
    - This is mainly done through nanostate, which provides basic finite state machines
    - As a result many things have better null states
    - Generally easier to read and understand
  * The only global events other than nanochoo's are `ws:*`, `switchserver`, `usesecure`, and `updatesession`
- Probably fixed a bunch of state-related/edge-case bugs we haven't noticed yet
- Fixed #167
- Fixed #139 
- Channel list now actually shows unread icons

To-do:
- Convert everything else (messages list (#137), emotes, account settings) to nanocomponents
- Fix #134 - already figured out a possible solution for this, just needs implementation. This is a longstanding issue and I'd like to fix it _now_ :package:
- `standard --fix` codestyle problems


packages/client/src/app.js
```diff
@@ -154,93 +140,54 @@ app.use((state, emitter) => {
...
-      <main>
-        ${accountSettings.component(state, emit)}
-      </main>
-    </div>`
+*/
+
+// connect websockets to global emitter. events:
+//  - ws:any:*     (events from any websocket)
+//  - ws:HOST:*    (events from websocket connected to HOST)
+//  - ws:active:*  (events from the current host's websocket)
```
towerofnix: *These* are super cool.

packages/client/src/components/sidebar/auth-state.js
```diff
@@ -0,0 +1,283 @@
...
+      machineState: this.machine.state
+    }
+
+    return shouldUpdate
+  }
+
+  // TODO
+  openAccountSettings() {
+    window.alert('Todo!')
+  }
```
towerofnix: (TODO)

packages/client/src/components/sidebar/auth-state.js
```diff
@@ -0,0 +1,283 @@
...
+      css`
+        #sidebar .auth-state.needs-auth {
+          color: var(--blue);
+          background: var(--gray-200);
+        }
+      `
+
+      return html`
+        <section class='auth-state needs-auth'>
+          <span class='msg'>needs auth</span>
```
towerofnix: Is this temporary? I'm pretty sure this should display the server authorization message.

packages/client/src/util/ws.js
```diff
@@ -83,11 +71,23 @@ class WS extends Nanobus {
...
 
-module.exports = WS
+module.exports = (host, useSecure) => {
+  // if there's already an open websocket connection to this host, use
+  // it instead of creating a new one
+  if (pool.has(host)) {
+    return pool.get(host)
+  }
+
+  const ws =  new WS(host, useSecure)
```
towerofnix: Two spaces? 👀

packages/client/src/components/sidebar/auth-state.js
heyitsmeuralex: Oh, yep!

heyitsmeuralex:
@towerofnix Looks like I snuck a linter past you :eyes:

```
> npm run test
```

I figured it'd be useful to go for `standard` sooner rather than later - only difference to our usual style is spaces before function argument lists: `function foo (bar) {}`

You may want to consider bringing this in cli and server too?

packages/client/src/util/ws.js
heyitsmeuralex:
```
> standard --fix
```

heyitsmeuralex:
@towerofnix I'm still considering using Preact or something over choo. Ugh, frontend is difficult.

PullJosh:
> I'm still considering using Preact or something over choo. Ugh, frontend is difficult.

So... I was an avid supporter of the framework ([p]react/vue/choo/etc) hype train, but all the newfangled JS stuff is definitely a ~little~ lot fatiguing.

Unfortunately, I don't have a solution. :P

heyitsmeuralex:
I don't think this is going to get finished anytime soon, at least by me. I want to use preact over [nano]choo - a lot of ideas from this branch can come with it, but not the commits :ok_man:

PR #194, opened 2018-02-03
Fix CORS for X-Session-ID

heyitsmeuralex:
Cherrypicked from the `nanocomponents` branch.

### DIFF
--- a/packages/server/api.js
+++ b/packages/server/api.js
@@ -613,7 +613,7 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
       // For people viewing the API manually for whatever reason to have
       // something to reference, we provide some info about the server itself.
       message: `This is a Decent chat server. See the repo for details.`,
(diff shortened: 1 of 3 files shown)


packages/server/api.js
```diff
@@ -613,7 +613,7 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
       // For people viewing the API manually for whatever reason to have
       // something to reference, we provide some info about the server itself.
       message: `This is a Decent chat server. See the repo for details.`,
-      repository: 'https://github.com/towerofnix/decent',
+      repository: 'https://github.com/decent-chat/decent',
```
heyitsmeuralex: Also fixes this.

towerofnix:
@heyitsmeuralex Is `app.use(['api/*', '/api'], ...)` in `api.js` necessary anymore with this change?

heyitsmeuralex:
I don't think so.

PR #195, opened 2018-02-03
Respond full user list if authorization is disabled

### LINKED ISSUE #191: /api/user-list doesn't work
Instead of returning a list of every user on the server, it returns an empty array.

towerofnix:
Fixes #191.

It turns out we weren't checking for authorization properly, so the "else" block wasn't being activated. And there was a typo in `runIfCondition` preventing "else" blocks from working, anyways; that's fixed as well.

@heyitsmeuralex I marked this semver: patch, but I'm not actually experienced with semantic versioning; is that right?

### DIFF
--- a/packages/server/api.js
+++ b/packages/server/api.js
@@ -341,7 +341,7 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
         }
       }).concat(runIfNot.map(callback => async (request, response, next) => {
         if (!await conditionFn()) {
-          callback(request, rseponse, next)
+          callback(request, response, next)
         } else {
           next()
         }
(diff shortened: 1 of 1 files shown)


packages/server/api.js
```diff
@@ -341,7 +341,7 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
         }
       }).concat(runIfNot.map(callback => async (request, response, next) => {
         if (!await conditionFn()) {
-          callback(request, rseponse, next)
```
heyitsmeuralex: gotta love js

packages/server/api.js
```diff
@@ -1225,7 +1225,7 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
   ])
 
   app.get('/api/user-list', [
-    ...middleware.runIfCondition(() => shouldUseAuthorization, [
+    ...middleware.runIfCondition(() => shouldUseAuthorization(), [
```
heyitsmeuralex: ditto

heyitsmeuralex:
As far as I can tell this should be patch.

Quick rundown:
* patch is for non-breaking patches that make the server follow the spec better
* minor is for backwards-compatible changes that don't fall under patch
* major is for breaking changes

Issue #173, opened 2018-01-14
Tests


heyitsmeuralex:
We need to create a test suite to cover the specification we have in docs/api.md.

towerofnix:
This should be done *soon*. I really want to figure it out.

A case of a dumb issue which definitely would have been caught if we'd written a test for it: #195.

PR #197, opened 2018-02-03
Add volatile in-memory database option

heyitsmeuralex:


PullJosh:
Is there any merit to disallowing this option during production? Just to help prevent anybody from doing something they didn't mean to do...

heyitsmeuralex:
@PullJosh it logs `Using in-memory database; image uploading is disabled` on start, so it shouldn't be too bad.

This is mainly for #173.

PullJosh:
> It shouldn't be too bad.

Your optimism is inspiring.

PR #201, opened 2018-02-03
Clean up API

### LINKED ISSUE #196: Clean up data returned from the API
The API is just a bit messy. My thoughts on the way it Should be:

* Scrap the `success` property. Stick to HTTP status codes, and...
* Standardize `error`s. The value of `error` should be an object, `error: {code, message}`, where `message` is a human-readable message (what we currently store in `error`) and `code` is a specific, permanent identifier string (e.g. `USER_NOT_FOUND`, `NONEXISTENT_SESSION`).
* Returned data should always be an object. This is so that we can attach "extra" data, e.g. making `user-list` return `users: {}` to everybody but also `unauthorizedUsers: {}` to admins. It also makes the return value a standard type, i.e. the object dictionary; we should never just return an array or string. Finally, it makes the return data be a *specific* object, and avoids key conflicts; for example, if we return `{users: {error: {id: 123}}}`, then `error` is a property of `<response>.users` and not of `<response>` (so it doesn't look like there's an actual error with the request; there's just a key on `users` named error).

The actual specification:

```
METHOD /api/any/endpoint
-> {
  // Present only if there was an error with the request:
  error: {
    code: 'SOME_ERROR_CODE',
    message: 'There was a very bad error.'
  },

  // Present only if there was no error (these are endpoint-specific):
  prop1: [123],
  prop2: 'blah'
}
```

towerofnix:
Fixes #196 by changing the API as described there. This is obviously a breaking change (the client is updated in this PR as well).

> This first pass focuses on two things:
>
> 1. Destroy any usage of `{success: true}`.
> 2. Make sure all returned values are an object, with specific details (e.g. user list) responded in separate properties (e.g. `"users": []`) and not stored on the top-level returned object itself.
>
> ---
>
> This part finishes up the clean-up, adding proper errors. Docs need to be updated.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -216,6 +216,26 @@ Deletes the given sessions (using their IDs will no longer work). Passing just o
 
 Returns `{sessions}` if successful, where `sessions` is an array of [(brief) session objects](#session-object). All sessions which are logged into the same user as the given session (via `sessionID`) are returned.
 
+
+## Endpoint errors
+
+Nearly all of the [HTTP endpoints](#http-endpoints) return errors situationally. The following list describes each:
+
+- `NOT_FOUND` - For when you try to request a something, but it isn't found (e.g. requesting the user by the name `foobar` when there is no such user).
+- `NOT_YOURS` - For when you attempt to do something impactful (e.g. editing/deleting) to a something you aren't the owner/author of.
+- `MUST_BE_ADMIN` - For when you try to do something limited to admins, but you are not an admin.
+- `ALREADY_PERFORMED` - For when you try to do something, but you have already done that something (e.g. pinning a message you've already pinned).
+- `INCOMPLETE_PARAMETERS` - For when a property is missing from a request's parameters. The missing property's name is passed in `error.missing`.
+- `INVALID_PARAMETER_TYPE` - For when a property is given in a request's parameters, but is not the right type (e.g. passing a string instead of an array). The invalid property's name is passed in `error.invalidParameter`.
+  - Note that this is only for type-checking. Client programs should *never* get this error, regardless of user input. More specific errors, such as `SHORT_PASSWORD`, are responded for issues that might be related to user input.
+- `INVALID_SESSION_ID` - For when a session ID is passed, but there is no session with that ID. (This is for general usage where being logged in is required. For `/session/:sessionID`, `NOT_FOUND` is returned if there is no session with the given ID.)
+- `UPLOAD_FAILED` - For when an upload fails.
+- `NAME_ALREADY_TAKEN` - For when you try to create a something, but your passed name is already taken by another something (e.g. registering a username which is already used by someone else).
+- `SHORT_PASSWORD` - For when you attempt to register but your password is too short.
+- `INCORRECT_PASSWORD` - For when you attempt to log in but you didn't enter the right password. (Note that `NOT_FOUND` is returned if you try to log in with an unused username.)
+- `INVALID_NAME` - For when you try to make something (a user or channel, etc) with an invalid name.
+
+
 ## WebSocket events
 
 These are the events which are used to send (and receive) data specific to individual connections to the server, and for "live" updates (e.g. rather than having the client poll the server for new messages every 5 seconds, the server emits a message to the client's web socket whenever a new message appears). These should certainly be considered when designing a custom client as they are the primary way in which the server talks to the client.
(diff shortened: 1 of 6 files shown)


docs/api.md
```diff
@@ -193,28 +193,48 @@ Returns `{available}`, where `available` is a boolean set to whether or not the
...
 
 - [Authentication](#authentication): always required.
 
-Returns `{success: true, sessions}` if successful, where `sessions` is an array of [(brief) session objects](#session-object). All sessions which are logged into the same user as the given session (via `sessionID`) are returned.
+Returns `{sessions}` if successful, where `sessions` is an array of [(brief) session objects](#session-object). All sessions which are logged into the same user as the given session (via `sessionID`) are returned.
+
+
+## Endpoint errors
+
+Nearly all of the [HTTP endpoints](#http-endpoints) return errors situationally. The following list describes each:
```
heyitsmeuralex: Could you describe the `{ error: { code, message }}` format here?

packages/server/api.js
```diff
@@ -487,7 +490,7 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
             // the username variable is passed. To make this a little less
             // evil, it's possible for that word to be passed manually
             // (as the second argument to requireNameValid).
```
heyitsmeuralex: Does this comment still apply?

packages/server/api.js
```diff
@@ -661,12 +664,12 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
     (req, res) => uploadSingleImage(req, res, err => {
       if (err) {
         res.status(500).end(JSON.stringify({
-          error: err.message
+          error: Object.assign(errors.UPLOAD_FAILED, {message: err.message})
```
heyitsmeuralex: `Object.assign` mutates its first argument, doesn't it? Fix pls.

packages/server/api.js
```diff
@@ -1384,7 +1360,7 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
 
       if (sessionUser._id === userID) {
         response.status(400).end(JSON.stringify({
-          error: 'you cannot deauthorize yourself'
+          error: errors.AUTHORIZATION_ERROR
```
heyitsmeuralex: Ehh... this is a mix between `NOT_YOURS` and `AUTHORIZATION_ERROR`, isn't it? Could you use a custom `message` here to clarify?

packages/server/api.js
```diff
@@ -1411,20 +1385,18 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
 
       if (Array.isArray(sessionIDs) === false) {
         response.status(400).end(JSON.stringify({
-          error: 'expected sessionIDs to be an array'
+          error: errors.INVALID_PARAMETER_TYPE
         }))
       } else if (sessionIDs.find(x => typeof x !== 'string')) {
         respones.status(400).end(JSON.stringify({
-          error: 'expected sessionIDs to be an array of strings'
+          error: errors.INVALID_PARAMETER_TYPE
```
heyitsmeuralex: Clarification of the expected type here too would be nice. We could have a `error.invalidParameterType(key, expected, actual)` for generating these..?

heyitsmeuralex:
Also note merge conflicts - there's a new error in there we need to add, notably `UPLOADS_DISABLED`.

heyitsmeuralex:
Also it'd be nice to catch internal errors and return code 500 and `{ error: { code: 'INTERNAL_ERROR', message: 'ReferenceError: ...' }}` or something.

packages/server/api.js
towerofnix: Nope, thanks. Removed (commit to be pushed).

Funnily enough, I don't think anything actually passed an `errorFieldName` to `requireNameValid`.

packages/server/api.js
towerofnix: Ooops! I meant to do `Object.assign({}, errors.UPLOAD_FAILED, {message: err.message})`. Thanks.

packages/server/api.js
towerofnix: Sure, I'll add a custom message. I didn't put too much thought into AUTHORIZATION_ERRORs because they're going to be obsolete pretty near in the future.

packages/server/api.js
towerofnix: I totally want an `invalidParameterType` function, but I think it's a liiiittle out of the scope of this PR.

Anyways, I'll add custom messages to these two.

towerofnix:
>Also note merge conflicts - there's a new error in there we need to add, notably UPLOADS_DISABLED.

Right, fixed. Also fixed the other things you mentioned and added info regarding what errors actually look like to the docs.

PR #204, opened 2018-02-05
API redesign + tests

heyitsmeuralex:


packages/server/middleware.js
```diff
@@ -307,7 +342,10 @@ module.exports.makeMiddleware = function({db, util}) {
 const validate = {
   string: Object.assign(function(x) {
     return typeof x === 'string'
-  }, {description: 'a string'})
+  }, {description: 'a string'}),
+  object: Object.assign(function(x) {
+    return typeof x === 'object'
```
heyitsmeuralex: `Object.is`

packages/server/test/api-messages.js
```diff
@@ -0,0 +1,36 @@
+const { test } = require('ava')
+const { makeChannel, makeUser, testWithServer } = require('./_serverUtil')
+const fetch = require('./_fetch')
+
+let portForApiMessageTests = 26000
+
+test('POST /api/messages', t => {
+  return testWithServer(portForApiMessageTests++, async ({ server, port }) => {
```
heyitsmeuralex: Don't know when you implemented this, but :+1:

packages/server/test/middleware.js
```diff
@@ -188,6 +188,12 @@ test('validate.string', t => {
   t.false(validate.string())
 })
 
+test('validate.object', t => {
+  t.true(validate.object({}))
+  t.false(validate.object(123))
+  t.false(validate.object())
```
heyitsmeuralex: `t.false(validate.object([])` thanks

heyitsmeuralex:
Note that with `response.json` you can `app.set('json spaces', 2)` to prettify responses. May be worth doing that when `process.env.NODE_ENV !== 'production'`!

packages/server/middleware.js
towerofnix: 👍, changed/tested!

packages/server/test/api-messages.js
towerofnix: I realized that our server-based tests weren't having their servers be killed by `server.kill()` because of the test failing before it reached that point. `testWithServer` gets around that with a `try`/`finally` block.

packages/server/api.js
```diff
@@ -234,6 +234,56 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
...
+    async function(request, response, next) {
+      const { imageURL, shortcode } = request[middleware.vars]
+
+      if (await db.emotes.findOne({shortcode})) {
+        response.status(400).json(errors.NAME_ALREADY_TAKEN)
+        return
+      }
+
+      const newEmote = {imageURL, shortcode}
+      await db.emotes.insert(newEmote)
```
heyitsmeuralex: Since it's unique anyway it might be worth using `_id` rather than `shortcode` on the db side. Small optimization, I guess.

packages/server/api.js
```diff
@@ -234,6 +234,56 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
...
+      sendToAllSockets('emote/new', {
+        emote: await serialize.emote(newEmote)
+      })
+      response.status(200).json({})
```
heyitsmeuralex: `201 Created` please. Same goes for `POST /api/users` and friends.

packages/server/api.js
```diff
@@ -234,6 +234,56 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
     ])
   }
 
+  app.post('/api/emotes', [
+    ...middleware.loadSessionID('sessionID'),
+    ...middleware.getSessionUserFromID('sessionID', 'sessionUser'),
+    ...middleware.requireBeAdmin('sessionUser'),
+    ...middleware.loadVarFromBody('shortcode'),
+    ...middleware.validateVar('shortcode', validate.string),
```
heyitsmeuralex: `validate.name`?

packages/server/test/api-emotes.js
```diff
@@ -0,0 +1,64 @@
+const { test } = require('ava')
+const { testWithServer, makeAdmin } = require('./_serverUtil')
+const fetch = require('./_fetch')
+const nodeFetch = require('node-fetch')
+
+let portForApiEmoteTests = 27000
```
heyitsmeuralex: Eh, what was the reasoning for removing `findport`? Sometimes ports can be used and this might make tests fail on particular machines - would be nice if `spawn` picked an open port at least (logic: `port++; if port.open return port else recurse`)

packages/server/api.js
towerofnix: Naw, I agree but this future-proofs it from any wild ideas we have in the future where we might want two emotes to have the same shortcode. No need to actively *prevent* that. What's going to take up *more* database space are individual messages themselves :)

packages/server/api.js
towerofnix: 👍

packages/server/test/api-emotes.js
towerofnix: See https://github.com/decent-chat/decent/pull/204/commits/93c6d481b6e15bbbdcb59097f9072fb19ea9c68f.

>```
>Unfortunately, openPort doesn't work well when we're dealing with a lot
>of things running concurrently (and in separate processes
>simultaneously). Instead, we just use a base port number constant in
>each test file and increase that for each spawned server. (So, 21000 in
>tests/auth.js, 22000 in tests/middleware.js, 23000 in the next test file
>we make, etc -- these need to be very far from each other because we
>don't want to have conflicts and, as I mentioned, openPort doesn't work.)
>```

I agree with you, though. Want to have a go at making that more clean and avoid used ports? I still think we should keep these ports separated in blocks of [big number, currently 1000] though - as it would turn out, there's sometime between when we decide a port number (e.g. via `openPort`) and when that port is actually used (probably because of databases opening, etc). And that delay becomes much more significant when we're dealing with two test files running literally concurrently.

docs/api.md
```diff
@@ -1,386 +1,840 @@
...
 
-### POST `/api/delete-sessions`
+<- {
+<-   "sessionID": "12345678-ABCDEFGH"
+<- }
+```
 
-- Parameters:
-  * `sessionIDs`: (via body; array of strings) the IDs of the sessions to be deleted.
+### Fetch session details [GET /api/session/:id]
```
towerofnix: I'm changing this to `/api/sessions/:id`, by the way, to match `/api/users/:id`, `api/channels/:id`, etc.

packages/server/api.js
```diff
@@ -384,39 +384,6 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
...
-        $push: {
-          pinnedMessageIDs: messageID
-        }
-      })
-
-      response.status(200).json({})
-    }
-  ])
-
   app.post('/api/add-message-reaction', [
```
heyitsmeuralex: Oh yeah, still need to do reactions. I'm thinking `POST /api/messages/reactions` to create, etc.?

packages/server/api.js
```diff
@@ -606,10 +610,11 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
     }
   ])
 
-  app.post('/api/delete-channel', [
-    ...middleware.loadVarFromBody('channelID'),
+  app.delete('/api/channels/:channelID', [
+    ...middleware.loadSessionID('sessionID'),
     ...middleware.getSessionUserFromID('sessionID', 'sessionUser'),
     ...middleware.requireBeAdmin('sessionUser'),
+    ...middleware.loadVarFromParams('channelID'),
```
heyitsmeuralex: Hey, should we just use `loadVar` which tries request.body and falls back to query params?

heyitsmeuralex: The spec doesn't specifically allow or disallow request bodies for DELETE requests, for example.

packages/server/api.js
towerofnix: Yes, but honestly I'm not worrying much about reactions right now. They 100% do not work with the emote system yet. So don't bother documenting reactions, either.

packages/server/api.js
towerofnix: Okay, that sounds reasonable.

packages/server/middleware.js
```diff
@@ -66,6 +66,15 @@ module.exports.makeMiddleware = function({db, util}) {
...
 
+    loadVarFromQueryOrBody: (key, required = true) => [
+      // Combines loadVarFromQuery and loadVarFromBody. The body recieves priority.
+
+      ...middleware.loadVarFromBody(key, false),
+      ...middleware.loadVarFromQuery(key, false), // Does not overwrite.
+
+      ...middleware.validateVar(key, required ? validate.defined : async x => true),
+    ],
+
```
towerofnix: This is actually a super clever implementation, cool.

You should add tests to middleware.js, by the way (just take a look at the other two for example). Remember to test priority, i.e. that body is prioritized over query, as well as the "required" argument!

heyitsmeuralex:
I honestly think we should defer updating the client code to #187 (or similar), since most of the code changes there anyway. This PR is huge as it is.

packages/server/middleware.js
```diff
@@ -0,0 +1,373 @@
...
+const validate = {
+  string: Object.assign(function(x) {
+    return typeof x === 'string'
+  }, {description: 'a string'}),
+  object: Object.assign(function(x) {
+    return typeof x === 'object' && !Array.isArray(x)
+  }, {description: 'an object'}),
+  defined: Object.assign(function(x) {
+    return typeof x !== 'undefined'
+  }, {description: 'defined'})
```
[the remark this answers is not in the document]
heyitsmeuralex: Yup. Perhaps `middleware.varMustExist` would work better.

docs/api.md
```diff
@@ -780,6 +780,21 @@ PATCH /api/users/1
 <- {}
 ```
 
+<a name='check-username-available'></a>
+### Check if a username is available [GET /api/username-available/:username]
```
heyitsmeuralex: Hey, we could maybe generalize this for other endpoints. `GET /api/{channels,users}/name-available/:name`

packages/server/api.js
```diff
@@ -891,8 +833,78 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
...
+        response.status(500).json({
+          error: errors.NAME_ALREADY_TAKEN
+        })
+
+        return
+      }
+
+      if (password.length < 6) {
+        response.status(400).json({
+          error: errors.SHORT_PASSWORD
```
heyitsmeuralex: `{ expectedLength: 6 }`?

docs/api.md
towerofnix: Ideally, yeah, but it'd be best to figure out how to first prevent nedb/shortid from generating `name-available` as an ID!

packages/server/api.js
towerofnix: Nah, that's already in the error message. We can make minimum required password length a setting in the future, but since it's hard-coded as 6 characters minimum, there's no need to add an extra `expectedLength` property.

packages/server/test/api-users.js
```diff
@@ -0,0 +1,192 @@
...
+    const response = await fetch(port, '/username-available/glory')
+    t.deepEqual(response, {available: true})
+
+    await makeUser(server, port, 'glory')
+
+    const response2 = await fetch(port, '/username-available/glory')
+    t.deepEqual(response2, {available: false})
+
+    try {
+      await fetch(port, '/username-available/burrito lord')
```
heyitsmeuralex: lul

docs/api.md
towerofnix: (Also `/api/roles/name-available` if we go with that system.)

heyitsmeuralex:
lgtm 🎉

PR #208, opened 2018-02-23
Fix client

### LINKED ISSUE #12: Pressing escape in modals should cancel
For example, click a message to edit it, then press escape.

### LINKED ISSUE #13: Holding alt and pressing the up/down keys should navigate channels
Ala Discord and Riot.

### LINKED ISSUE #139: Don't do anything when the selected sidebar is clicked again

towerofnix:
The client used to be dead. Now it isn't.

This changes old endpoints in the client's code to new endpoints. I probably missed some, but this is a good start.

Some other things I changed:

* You'll automatically be logged in after registering.

* You'll automatically get kicked out of the account settings page if you log out while in it (e.g. by deleting your current session, or just clicking the "logout" button).

* Holding alt while pressing the up/down arrow keys switches to the channel above or below the current one in the channel list. This loops - if you're on the top channel and press up, it'll move to the bottom one (and vice versa). (Fixes #13.)

* Viewing a channel immediately focuses the message textarea (if you're logged in).

* Pressing escape cancels the active modal. (Fixes #12.)

* Clicking the channel you're already viewing won't reload the channel anymore. (Fixes #139.)

* No more "reset password" section in the account settings. It's always shown a "not implemented" alert, and it's really not necessary to leave it or its code lying around. It's not hard to recover the code when we eventually implement resetting passwords.

* The image upload API works now. I had to change some things on the backend.

* The account settings API also works. More (small) backend changes.

* Pages under `/api/` don't return the client's index.html if they aren't found. (Instead, they return the Express default - which is an HTTP-404 with a "Cannot GET /api/(...)" message).

* I removed a few redundant `sessionID`s being passed in the client. `util/api` automatically adds the session ID.

I've only tested this on a server which does not require authorization.

### DIFF
--- a/packages/client/src/components/message-editor.js
+++ b/packages/client/src/components/message-editor.js
@@ -14,10 +14,8 @@ const component = (state, emit) => {
     if (text.length === 0) return
     textarea.value = ''
 
-    await api.post(state, 'send-message', {
-      text,
-      channelID: state.params.channel,
-      sessionID: state.session.id,
+    await api.post(state, 'messages', {
+      text, channelID: state.params.channel,
     })
   }
 
@@ -57,15 +55,15 @@ const component = (state, emit) => {
     progressBar.classList.add('moving')
 
     try {
+      console.log(state.session.id)
       const { path } = await api.postRaw(state, 'upload-image?sessionID=' + state.session.id, formData)
 
       progressBar.style.width = '90%'
 
       // send a message with the image in it
-      await api.post(state, 'send-message', {
+      await api.post(state, 'messages', {
         text: `![](${state.secure ? 'https' : 'http'}://${state.params.host}${path})`,
-        channelID: state.params.channel,
-        sessionID: state.session.id,
+        channelID: state.params.channel
       })
 
       progressBar.style.width = '100%'
--- a/packages/client/src/components/message-editor.js
+++ b/packages/client/src/components/message-editor.js
@@ -20,10 +20,10 @@ const component = (state, emit) => {
   }
 
   textarea.addEventListener('keydown', evt => {
-    const key = evt.which
+    const key = (evt.which || evt.keyCode)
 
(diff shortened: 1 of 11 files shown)


PullJosh:
This sounds awesome! I'll try it out when I can. :)

heyitsmeuralex:
Returning { error: { code: "NOT_FOUND" } } would be fun over express defaults.

towerofnix:
@heyitsmeuralex Right, that's changed now.

packages/client/src/components/message-editor.js
```diff
@@ -91,6 +101,12 @@ const component = (state, emit) => {
       return a.className === editor.className
     }
 
+    // Hack!!! - Select the textarea *soon*. We assume that the component is
+    // rendered and on the page by 25ms from now.
+    setInterval(() => {
+      textarea.focus()
+    }, 25)
```
heyitsmeuralex: oof

packages/client/src/util/api.js
```diff
@@ -65,6 +79,22 @@ module.exports = {
     })
   },
 
+  patch(state, path, changes = {}) {
+    // PATCH takes a body, like POST.
+
+    if (state.session.id && !data.sessionID) {
+      data.sessionID = state.session.id
```
heyitsmeuralex: We should be using `X-Session-ID` everywhere imo.

heyitsmeuralex: You're free to merge if you want to fix this later, @towerofnix.

towerofnix: Agreed, changed that.

PR #209, opened 2018-02-23
User list

### LINKED ISSUE #68: User list UI

towerofnix:
Demo screenshot

Fixes #68.

User list is sorted by prioritizing online users, then alphabetically (A at top).

See also [PullJosh's UI reference](https://codepen.io/PullJosh/pen/OOYZWK). Most new CSS is copied from there, and I reused a lot of the sidebar's CSS (just by including it as another class prefix, not by copy/pasting it!).

@heyitsmeuralex Not sure what semver this is?

### DIFF
--- a/packages/client/src/app.js
+++ b/packages/client/src/app.js
@@ -14,6 +14,7 @@ Object.assign(window, { util })
 const messages = require('./components/messages')
 const messageEditor = require('./components/message-editor')
 const sidebar = require('./components/sidebar')
+const userList = require('./components/user-list')
 const accountSettings = require('./components/account-settings')
 const srvSettings = {
   emotes: require('./components/srv-settings/emotes'),
@@ -152,6 +153,7 @@ app.use((state, emitter) => {
 
 app.use(messages.store)
 app.use(sidebar.store)
+app.use(userList.store)
 app.use(accountSettings.store)
 
 for (const [ name, s ] of Object.entries(srvSettings)) {
@@ -186,6 +188,15 @@ for (const [ name, s ] of Object.entries(srvSettings)) {
     </div>`
   })
 
+  // server
+  app.route('/servers/:host', (state, emit) => {
+    return html`<div id='app'>
+      ${sidebar.component(state, emit)}
+      <main></main>
+      ${userList.component(state, emit)}
+    </div>`
+  })
+
   // server with channel open
   app.route('/servers/:host/channels/:channel', (state, emit) => {
     return html`<div id='app'>
@@ -194,6 +205,7 @@ for (const [ name, s ] of Object.entries(srvSettings)) {
         ${messages.component(state, emit)}
         ${state.messages.list !== null ? messageEditor.component(state, emit) : html`<span></span>`}
       </main>
+      ${userList.component(state, emit)}
     </div>`
   })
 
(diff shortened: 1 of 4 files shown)


heyitsmeuralex:
If we're going to make offline users opaque, the online/offline indicator circle doesn't really have a purpose anymore. Might I suggest that gets removed?

heyitsmeuralex:
Semver for this would be minor, because it **adds functionality in a backwards-compatible manner**.

packages/client/src/app.js
```diff
@@ -54,13 +55,16 @@ app.use((state, emitter) => {
 
   state._session = new Proxy({}, {
     set: function(target, key, value) {
-      if (key === 'id') {
+      const ret = Reflect.set(target, key, value)
```
heyitsmeuralex: Woah, what does this do? ...returns a boolean if `target[key] = value` was successful...? wat?

Conclusion: Proxy and Reflect are both magic. I realise this existed before but never noticed it's usage. thisandagain pls explain.

packages/client/src/app.js
```diff
@@ -152,6 +156,7 @@ app.use((state, emitter) => {
 
 app.use(messages.store)
 app.use(sidebar.store)
+app.use(userList.store)
 app.use(accountSettings.store)
```
heyitsmeuralex: aaaa I should really get started on fixing data flow in the app because this is bad

towerofnix:
Sounds good, removed the bubble.

packages/client/src/components/user-list.js
```diff
@@ -0,0 +1,116 @@
...
+              return +1
+            }
+
+            if (a.username < b.username) {
+              return -1
+            }
+
+            if (b.username < a.username) {
+              return +1
+            }
```
heyitsmeuralex: Could be worth formatting this as

```js
.sort((a, b) => do {
  if (cond) -1
  if (cond) +1
  0
})
```

Wait, no, `do {}` isn't in Stage 3 yet. Nevermind.
If we ever end up doing babel - probably when I replace choo with preact - using its do expression transform would be awesome (especially for (JSX/bel)related things!)

packages/client/src/components/user-list.js
```diff
@@ -97,7 +97,10 @@ const component = (state, emit) => {
             return 0
           }).map(user => {
             return html`
-              <div class='item user ${user.online ? 'online' : 'offline'}'>
+              <div
+                class='item user ${user.online ? 'online' : 'offline'}'
+                title='${user.username} (${user.online ? 'Online' : 'Offline'})'
```
heyitsmeuralex: :heart:

towerofnix:
I'm actually wondering if they maybe should be kept, though -- profile pictures which inherently aren't that saturated can look like they're offline, and then the only way to tell whether someone's online or not is by checking how dark their username text is.

towerofnix:
Comparison screenshot:

heyitsmeuralex:
Honestly I think the opacity of the username itself gives it away quite clearly, especially due to how the list is sorted.

towerofnix:
Alright! If it's a huge problem in the future we can always change it back.

heyitsmeuralex:
@towerofnix just added issue label descriptions for everything including the `semver` ones, so you should have a slightly less difficult time figuring out if it's major, minor, or a patch.

towerofnix:
@heyitsmeuralex Huh, cool! I didn't know label descriptions were a thing. Definitely handy.

heyitsmeuralex:
Pretty sure they're a very new thing.

PR #212, opened 2018-02-24
Fix authorization-related errors

towerofnix:
None of these are tracked, but this commit fixes authorization errors on both the client and the server.

It's a major/breaking change for the server (see below), and a patch change for the client (since it just fixes bugs).

* Fixes properties in general not being fetched correctly (on the client). `state.secure`, `state.serverRequiresAuthorization`, and `state.authorizationMessage` now actually work.

* Fixes the user list causing several errors that prevent messages from being loaded and such (mainly because it assumed a non-authorized server).

* Changes PATCH /api/settings to POST /api/settings (this is the breaking change, see below).

Important commit message:

>For some reason, no matter absolutely what, Express wasn't handling PATCH /api/settings. I have no idea why. Changing the method to POST /api/settings fixed it.
>
>I also updated the documentation, now also including information on the "results" return property.

### DIFF
--- a/packages/client/src/util/api.js
+++ b/packages/client/src/util/api.js
@@ -27,6 +27,14 @@ async function fetchHelper(state, path, fetchConfig = {}) {
   // if we get an error object, throw
   if (result.error) {
     // { message, data }
+
+    /*
+    console.log('error ---- ' + result.error.code)
+    console.log('fetch config:', fetchConfig)
+    console.log('sessionid:', state.session.id)
+    console.log('path:', path)
+    */
+
     throw Object.assign(new Error(result.error.message), {
       code: result.error.code,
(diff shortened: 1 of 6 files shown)


packages/client/src/util/api.js
```diff
@@ -27,6 +27,14 @@ async function fetchHelper(state, path, fetchConfig = {}) {
...
   if (result.error) {
     // { message, data }
+
+    /*
+    console.log('error ---- ' + result.error.code)
+    console.log('fetch config:', fetchConfig)
+    console.log('sessionid:', state.session.id)
+    console.log('path:', path)
+    */
+
```
towerofnix: I intentionally left this here for future debugging.

PR #225, opened 2018-02-26
Use uglify-es instead of tinyify

heyitsmeuralex:
Fixes build errors related to tinyify/uglify that kept on occurring, namely on mercury - the https://meta.decent.chat production server!! - and, sometimes, my local machine.

Oh, and I removed `dependency-check` and the `__kb` display after builds too because we don't really need them and nobody likes bloat.

CC @TheInitializer - this removes the `rm dist/*;` part of the build step so it should start working on Windows now, please test.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -22,18 +22,15 @@
   "devDependencies": {
     "browserify": "^15.1.0",
     "css-extract": "^1.2.0",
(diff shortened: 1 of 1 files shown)


TheInitializer:
@heyitsmeuralex yep, it works on windows! 🎉

Issue #185, opened 2018-01-15
Dark theme / CSS refactor


heyitsmeuralex:

heyitsmeuralex:
Note #187 removes the `:host` class prefix stuff in favour of static classnames (e.g. `._xyz123` -> `.channel-list`), making theming a lot more viable.

towerofnix:
Generally, I think it would make sense to refactor the CSS to use more descriptive variable names, e.g. "pane background". For example:

```css
body {
  --gray-900: #fff;
  --pane-background: var(--gray-900);
}

.sidebar (or :host, whatever) {
  background: var(--pane-background);
}
```

PullJosh:
While we're already looking at refactoring CSS, would there be any benefit to switching to [BEM](http://getbem.com/introduction/)?

heyitsmeuralex:
CSS has now been refactored. Only thing left is to make a dark theme or theming UI in the client.

towerofnix:
Implemented in #250.

PR #226, opened 2018-02-26
Refactor CSS

heyitsmeuralex:
See #185. This doesn't actually add a dark theme, but gets us 85% of the way there. Work-in-progress.

~~Blocked by #225 as this includes it.~~


heyitsmeuralex:
Requested a review from @towerofnix as the rest of the work is just refactoring. Would like an opinion on the style etc.

heyitsmeuralex:
Only thing left is to make more prescriptive variables as discussed in #185.

packages/client/css/app.css
```diff
@@ -0,0 +1,106 @@
...
+
+a:any-link.link, .Link {
+  color: var(--blue);
+  text-decoration: underline;
+}
+
+::-moz-selection { background: #b3d4fc; }
+::selection { background: #b3d4fc; }
+
+table, th, td {
```
PullJosh: Feeling a little nervous about such generic selectors. Which tables are these styles meant to apply to? Where do they appear?

packages/client/css/generic/button.css
```diff
@@ -0,0 +1,28 @@
+.Button {
+  padding: 10px 16px;
+  &:not(:first-child) { margin-left: 3px; }
+  &:not(:last-child) { margin-right: 3px; }
```
PullJosh: Adding negatives (`:not`) tends to make things more confusing. Might it be better to say the following?

```css
margin-right: 6px;
&:last-child { margin-right: 0; }
```

This, I believe, would achieve the same effect: a `6px` gap between each `.Button`.

packages/client/css/generic/input.css
```diff
@@ -0,0 +1,66 @@
...
+
+    font-weight: bold;
+    font-size: 16px;
+
+    text-transform: uppercase;
+    margin-bottom: 4px;
+    margin-left: 13px;
+
+    -webkit-touch-callout: none;
+    -webkit-user-select: none;
```
PullJosh: Indentation doesn't quite line up. This problem probably also existed in the old CSS, but it should be fixed while we're in the process of reworking things.

packages/client/css/generic/input.css
```diff
@@ -0,0 +1,66 @@
...
+  background: var(--gray-900);
+
+  border: 0;
+  border-bottom: 1px solid var(--gray-500);
+  &:focus { border-color: var(--blue); }
+
+  &[disabled] {
+    cursor: not-allowed;
+    background: var(--gray-700);
+    opacity: .8;
```
PullJosh: There seems to be some inconsistency with whether or not to include a leading 0. We should choose a standard (my vote is for including the 0) and mention it in CONTRIBUTING.md.

CONTRIBUTING.md
```diff
@@ -75,3 +75,77 @@
...
+    &.is-loading {
+      /* ... */
+    }
+  }
+</style>
+
+<div class='Image is-loading'></div>
+```
+
+* Use `&-childName` for children
```
PullJosh: It seems that child component names begin with a lowercase, while parent components begin with an uppercase. Is this intentional? More importantly, is it a good idea?

packages/client/css/generic/message-group.css
```diff
@@ -0,0 +1,37 @@
...
+.MessageGroup-authorUsername {
+  display: inline-block;
+  margin-right: 4px;
+
+  font-weight: bold;
+  text-align: right;
+}
+
+.MessageGroup-contents {
+  width: calc(100% - 48px - 16px);
```
PullJosh: What is the reasoning for these numbers? I love separating out `48px` and `16px` (as opposed to `100% - 32px`) if they have distinct meanings, but it's not clear from the code what those meanings are.

A simple comment should clear things up.

packages/client/css/generic/message.css
```diff
@@ -0,0 +1,58 @@
...
+
+  object-fit: contain;
+  width: 1rem;
+
+  margin: 0 .05em 0 .1em;
+  vertical-align: -.4em;
+}
+
+/* note: this animation name is used in the JavaScript code,
+   so take care if renaming it */
```
PullJosh: :+1: This comment will save somebody one day.

packages/client/css/pages/account-settings.css
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+.AccountSettings-submit {
+  display: block;
+
+  margin-top: 80px;
+  margin-left: auto;
+  margin-right: 0;
+  margin-bottom: 20px;
```
PullJosh: Any reason not to consolidate these into a single `margin`?

packages/client/css/specific/modal.css
```diff
@@ -0,0 +1,76 @@
...
+  left: 50%;
+  transform: translate(-50%, -50%);
+  max-width: calc(100% - 48px);
+
+  background: var(--gray-900);
+  border-radius: 4px;
+  overflow: hidden;
+
+  z-index: 950;
+  box-shadow: 0 1px 8px #00000044;
```
PullJosh: 8-digit hex notation [doesn't seem](https://caniuse.com/#feat=css-rrggbbaa) particularly well supported. I know that browser support isn't priority numero uno, but it seems no worse to use `rgba(0, 0, 0, 0.27)` in this case.

packages/client/css/specific/modal.css
```diff
@@ -0,0 +1,76 @@
...
+  top: 16px;
+  right: 16px;
+  width: 19px;
+  height: 19px;
+  background: url('/img/x.svg') center center / contain;
+  cursor: pointer;
+}
+
+.Modal-title {
+  padding: 69px;
```
PullJosh: classic

packages/client/css/app.css
towerofnix: The only tables in use are currently in the emote and authorized users setting pages, I think.

packages/client/css/generic/input.css
towerofnix: My vote is 1000% for including the zero.

packages/client/css/generic/message-group.css
towerofnix: Gosh, yeah. @heyitsmeuralex I absolutely can't stand using `calc`. Could you rework this with flexbox, or even just float?

packages/client/css/pages/account-settings.css
towerofnix: (Note the order would be `margin: 80px 0 20px auto`.)

CONTRIBUTING.md
```diff
@@ -75,3 +75,77 @@
 +   3,
   ]
 ```
+
+## CSS Naming Conventions
+
+This is essentially our own flavour of BEM.
+
+* Use `.ComponentName` for components
```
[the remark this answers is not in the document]
heyitsmeuralex: Yeah. `Use pascal-case...`

packages/client/css/app.css
heyitsmeuralex: Yeah, these are leftovers! We should move them to more specific places.

packages/client/css/generic/input.css
heyitsmeuralex: wrong

CONTRIBUTING.md
heyitsmeuralex: Yeah, it's intentional. I think it makes it more obvious what the parent and child are.

packages/client/css/generic/message-group.css
heyitsmeuralex: aaaa No idea what this line is for :package:
I'll look into it.

packages/client/css/generic/message.css
heyitsmeuralex: `git blame` @towerofnix :star:

heyitsmeuralex:
Made all those requested changes, @PullJosh. Only thing left is `--prescriptive-variable-names`.

heyitsmeuralex:
Made all those requested changes, @PullJosh. Only thing left is `--prescriptive-variable-names`.

PR #228, opened 2018-02-27
Handle unread notifications on client-side

heyitsmeuralex:
Adds the necessary code to assume the state of `channel.unreadMessages` based on message WS events. Also calls `mark-as-read` when switching to a channel.

### DIFF
--- a/packages/client/src/components/sidebar.js
+++ b/packages/client/src/components/sidebar.js
@@ -199,6 +199,12 @@ const store = (state, emitter) => {
     // Don't switch to the channel if we're already viewing it!
     if (!(state.route === '/servers/:host/channels/:channel' && state.params.channel === id)) {
       emitter.emit('pushState', `/servers/${state.params.host}/channels/${id}`)
+
+      // Mark the channel as read
+      const channel = state.sidebar.channels.find(channel => channel.id === id)
+      channel.unreadMessageCount = 0
+
+      api.post(state, `channels/${id}/mark-read`)
     }
   })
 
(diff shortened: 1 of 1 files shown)


packages/client/src/components/sidebar.js
```diff
@@ -199,6 +199,12 @@ const store = (state, emitter) => {
     // Don't switch to the channel if we're already viewing it!
     if (!(state.route === '/servers/:host/channels/:channel' && state.params.channel === id)) {
       emitter.emit('pushState', `/servers/${state.params.host}/channels/${id}`)
+
+      // Mark the channel as read
+      const channel = state.sidebar.channels.find(channel => channel.id === id)
+      channel.unreadMessageCount = 0
+
+      api.post(state, `channels/${id}/mark-read`)
```
heyitsmeuralex: Is this needed? I'm not sure what the conditions are on the server for marking a channel is read without this endpoint.

iirc it was after a message is sent, which is not really what we want at the moment. Once we begin loading messages _from_ the oldest unread message when viewing a channel we'd only want to send this once the user reached the bottom of the messages (or when they click `Mark as read`).

towerofnix: @heyitsmeuralex Yes, the only time when a channel is automatically marked as read is when a message is sent (ctrl-F usages of `markChannelAsRead` in api.js).

> Once we begin loading messages *from* the oldest unread message...

I agree, that's how it should be implemented later. For now, since viewing a channel simply fetches the most recent messages (rather than the oldest unread messages), we may as well just call mark-read as soon as the channel is viewed, as you've done in this PR.

towerofnix: @heyitsmeuralex Yes, the only time when a channel is automatically marked as read is when a message is sent (ctrl-F usages of `markChannelAsRead` in api.js).

> Once we begin loading messages *from* the oldest unread message...

I agree, that's how it should be implemented later. For now, since viewing a channel simply fetches the most recent messages (rather than the oldest unread messages), we may as well just call mark-read as soon as the channel is viewed, as you've done in this PR.

PR #236, opened 2018-02-28
Add bigmoji

heyitsmeuralex:
Conditions for bigmoji:

* Message only contains emotes
* Max 16 emotes in a single message

### DIFF
--- a/packages/client/css/generic/message.css
+++ b/packages/client/css/generic/message.css
@@ -41,6 +41,10 @@
   vertical-align: -.4em;
 }
 
+.Message.\--big-emotes .Message-emote {
+  width: 4rem;
+}
+
 /* note: this animation name is used in the JavaScript code,
    so take care if renaming it */
 @keyframes jumped-message {
--- a/packages/client/src/components/message-group.js
+++ b/packages/client/src/components/message-group.js
@@ -45,8 +45,11 @@ const component = (state, emit, group) => {
       </div>
 
       ${group.messages.map(msg => {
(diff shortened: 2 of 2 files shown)


packages/client/css/generic/message.css
```diff
@@ -41,6 +41,10 @@
   vertical-align: -.4em;
 }
 
+.Message.\--big-emotes .Message-emote {
+  width: 4rem;
```
PullJosh: Wouldn't it make more sense to use `em` here, so that if we change the font size of the messages, the size of emojis changes too?

heyitsmeuralex: Yeah. Normal emotes use `rem` too, however, so this is probably out-of-scope for this PR.

packages/client/src/components/message-group.js
```diff
@@ -45,8 +45,11 @@ const component = (state, emit, group) => {
       </div>
 
       ${group.messages.map(msg => {
-        const el = html`<div class='Message' id=${'msg-' + msg.id}>
-          ${raw(mrk(state)(msg.text).html())}
+        const mrked = mrk(state)(msg.text)
+        const bigEmotes = mrked.tokens.length <= 16 && !mrked.tokens.find(t => t.name !== 'emote')
```
towerofnix: This is a pretty cool way of using `mrk`.

heyitsmeuralex: Yep! Part of the reason for its existence is access to the token stream.

heyitsmeuralex:
> Error 503 Backend is unhealthy

PR #237, opened 2018-02-28
Add preferences

heyitsmeuralex:
* Adds preferences section to sidebar
* Adds _Looks_ preferences page, with a single option - circular avatars via CSS variables - towards #185
* DRYified some `app.js` code
* Moved the previously-`.AccountSettings-input`-specific horizontal input style to `Input --horizontal`

image
image
image


PullJosh:
Nice! But can we please call it "appearance" instead? :package:

PR #241, opened 2018-02-28
Flairs

### LINKED ISSUE #163: Flairs
Reddit-style; so, for example, "John Smith [sysadmin]" or "GodBot [bot]". (These would just be tag-like strings that display beside a username in message headings. Might want to make them colorable?)

I also want to consider *roles*. Maybe roles could simply have a "show as flair" option? That would simplify things (and also let users have more than one flair).

### LINKED ISSUE #223: Changed profile pictures should reflect in user list sidebar
At the moment, the old profile picture will be used until the page is reloaded.

Thanks for spotting this, @eq!

towerofnix:
Implements flairs and adds a new `PATCH /api/users/:userID` endpoint, which replaces `POST /api/account-settings`, `POST /api/deauthorize-user`, and `POST /api/authorize-user`.

Resolves #163; fixes #223.

Flair demo


heyitsmeuralex:
OOF that's a lot of commits!

RE: passwords and salts; bcrypt is one of the only formats to store data like that in the hash itself, so don't worry.

heyitsmeuralex:
Note that this removes /api/(de)authorize-user!

towerofnix:
@heyitsmeuralex This is a minor change to the client, isn't it?

heyitsmeuralex:
Uhhhh.... yes. But because the API spec/server version would be bumped majorly by this and the client is updated to support it, they would all have the same version.

We're still sub-1.0.0, so we're not actually applying SemVer yet. We should make a 1.0.0 milestone, actually.

heyitsmeuralex:
Awaiting @towerofnix's review - it's your PR, so just add comments?

PR #251, opened 2018-03-01
Change <br> to <br/>

towerofnix:
Firefox does *not* like it when you try to `createElement('br/')`. This is admittedly a problem that should be fixed in choo, but it's easy enough to quickly change here.


heyitsmeuralex:
Also don't you mean `change <br/> to <br>`? :package:

PR #254, opened 2018-03-01
Add channel.oldestMessageID

heyitsmeuralex:
Towards loading channels from the oldest unread message (is there an issue for this?).

Blocked/includes #253.

Issue #115, opened 2017-12-17
Edit/delete messages


towerofnix:
Each message should have a way to open a dropdown (probably an ellipsis-button to the side) letting the user edit or delete it.

* Edit (if the user owns the message). See #58 for old discussion on how editing a message should work. @PullJosh? On Gitter and Discord, editing a message replaces the message's content with an input box for typing in; that seems the most intuitive to me. (Also, it doesn't remove existing content from the chat input - it creates a *new* input.)

* Delete (if the user owns the message, or is an admin). Should there be a confirmation dialog?

The actual dropdown should be implemented as a "separate" unit - other functions should be able to easily use the same dropdown code. I'm thinking there should be a global dropdown element which is filled/shown by a `showDropdown` function, that way we don't need to create and keep track of tooooooooooons of DOM elements (a dropdown and its items for every single message sent!).

PullJosh:
A few thoughts:
* Should there be checkboxes next to messages for admins so that they can mass-delete messages?
* The ideal UI for deleting messages would be to have a button which performs the action instantly (no confirmation modal), but then gives you an option to undo (a la gmail). No idea whether this is feasible API-wise though.

heyitsmeuralex:
>Should there be checkboxes next to messages for admins so that they can mass-delete messages?

Eh, that doesn't seem fun. We can do shift-clicking on messages to select, though - I feel like that's better than a checkbox.

heyitsmeuralex:
>but then gives you an option to undo. No idea whether this is feasible API-wise though

If we just set a flag (`{ deleted: true }`) on a message when we delete it, we can keep all metadata about that message. The server should just never send messages where `deleted: true` to clients.

PullJosh:
Would there be any benefit to creating two different options: "delete" and "hide"?

Delete could require a modal confirmation and be permanent, while "hide" just triggers a "hidden" boolean which makes the message invisible to regular users and collapsed and/or faded out for admins.

No idea if that would be better, just something that came to mind.

heyitsmeuralex:
I don't think that's worth it.

towerofnix:
Mind you, I want there to be *some* means of permanently deleting messages - as in, actually removing them from the database. Leaving them there (even if they're only internally stored and can't be fetched) seems like a breach of trust in what "deleting" means.

One option is to have a set amount of time after which messages are permanently deleted (server setting/property).

heyitsmeuralex:
We'd want to delete old image uploads too.

[...]

PR #255, opened 2018-03-01
Implement pinned messages + message deletion

### LINKED ISSUE #17: Pinned messages

heyitsmeuralex:
@PullJosh did a design for tabs in the right sidebar; this is it in fruition!

* Contains some API spec changes: two new events and an unpin endpoint.
* You can delete/pin/unpin messages now with the UI (see #115, fixes #17)
* Note that the `mentions` tab is left blank for when #92 is finally complete, and the edit message icon does nothing when clicked; see #115.
* Includes / iterates on #254 - channels are now loaded from the oldest unread message
* ~~Blocked by / includes #253. Merge that first.~~ :ballot_box_with_check:


towerofnix:
**Pre-code-review notes:**

Note - trying to view the "pins" tab while no channel is selected causes an error, because it fetches `/api/channels/undefined/pins`. Pins tab should probably be hidden while not actively in a channel? (And should switch to "users" tab when going back to a state with no channel active.)

Note - it's sorta hard to tell exactly what message my mouse is on top of, since the icons appear so far to the right. Maybe a dotted line from the message icons to the first line of the related message would help?

Note - on an authorized server, when I reloaded the page, there was an "authorization related error", but I'm 60% sure that doesn't have anything to do with this.

Note - flairs don't show up on pinned messages. I assume this is intentional? (It looks better without flairs imo.)

Note - the most recently pinned messages show up at the bottom of the pinned message list. IMO there should be a way to sort pinned messages - maybe even manually. Otherwise, they should probably be sorted by date. (Should the newest messages go on the top or the bottom?)

Note - the highlight from jumping to a pinned message doesn't look any different in the dark mode, but this is probably fine.

docs/api.md
```diff
@@ -475,18 +475,27 @@ Model:
...
+  "unreadMessageCount": number,
+
+  "oldestUnreadMessageID": ID | null,
 }
 ```
 
 Related events:
 * [channel/new](#channel-new)
 * [channel/update](#channel-update)
+* [channel/pins-update](#channel-pins-update)
```
towerofnix: This event does not exist.

docs/api.md
```diff
@@ -661,7 +670,7 @@ GET /api/channels/5678/pins
 + **in-url** id (ID)
 + `messageID` (ID) - The message to pin to this channel.
 
-Returns `{}` if successful.
+Returns `{}` if successful. Emits [channel/pins-update](#channel-pins-update).
```
towerofnix: This should be `channel/pins/add`.

packages/server/serialize.js
```diff
@@ -52,6 +52,15 @@ module.exports = function makeSerializers({util, db}) {
 
       if (sessionUser) {
         obj.unreadMessageCount = await getUnreadMessageCountInChannel(sessionUser, c._id)
+
+        if (obj.unreadMessageCount === 0) {
+          obj.oldestUnreadMessageID = null
+        } else {
+          const msg = await getOldestUnreadMessageInChannel(sessionUser, c._id)
+
+          console.log(msg)
```
towerofnix: Leftover console.log?

docs/api.md
towerofnix: (Replace this with channel/pins/add and channel/pins/remove).

heyitsmeuralex:
Hold on, just adding that dotted line thing you asked for.

PR #256, opened 2018-03-01
Pad messages more

heyitsmeuralex:
image

Suggested by @TheInitializer - been on my mind for a while but never got around to it. Separate messages now actually look like separate messages.


towerofnix:
Hold on, let me take a look before merging too :)

packages/client/css/generic/message.css
```diff
@@ -1,6 +1,6 @@
 .Message {
   line-height: 1.4;
-  margin-bottom: 1em;
+  margin-bottom: .25em;
```
towerofnix: Should have a leading zero.

heyitsmeuralex: That was fast. Force-pushed over it.

PR #258, opened 2018-03-02
Mentions

### LINKED ISSUE #92: Mentions

### LINKED ISSUE #227: Sidebars should be scrollable
Who wants to CSS position-sticky headers like "users" and "channels", which is now [supported on everything](https://developer.mozilla.org/en-US/docs/Web/CSS/position#Browser_compatibility)? Or maybe that's a dumb idea :P

Blocked by #226, but could be implemented into that (ping @heyitsmeuralex, feel free to unassign yourself if you don't want to do this in that PR).

heyitsmeuralex:
<sup>"mention" doesn't look like a word anymore <sup>help</sup></sup>

Adds mentions; fixes #92. Includes respective APIs and events (similar to the pins-related ones). Mentions are formatted as `<@user_id>` and are tracked by the server :package: 

Also:
* fixes #227 
* updated [mrk](/heyitsmeuralex/mrk) to `2.0.0` - it's now factory-function based and supports a `require('mrk.js/async')` Promise mode. We currently can't do async rendering so it's not very useful.


docs/api.md
```diff
@@ -713,8 +716,10 @@ Model:
 
   "online": boolean,
 
+  "mentions": [ Message ] // List of messages that mention this user
```
towerofnix: Missing comma.

packages/server/api.js
```diff
@@ -50,6 +51,56 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
     }
   }
 
+  const handleMentionsInMessage = async function (message) {
+    for (let userID of new Set(await getMentionsFromMessageContent(message.text))) {
```
towerofnix: Should be `for (const userID`, not `let`.

packages/server/api.js
```diff
@@ -50,6 +51,56 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
...
+      }
+    }
+  }
+
+  const handleUnmentionsInMessage = async function (message, newText) {
+    const mentionedOld = await getMentionsFromMessageContent(message.text)
+    const mentionedNew = await getMentionsFromMessageContent(newText)
+    const unmentionedInNew = mentionedOld.filter(id => !mentionedNew.includes(id))
+
+    for (let userID of unmentionedInNew) {
```
towerofnix: Again, should be `for (const of`.

packages/server/api.js
```diff
@@ -467,6 +520,8 @@ module.exports = async function attachAPI(app, {wss, db, dbDir}) {
         return
       }
 
+      await handleUnmentionsInMessage(oldMessage, text)
```
towerofnix: Does this remove old mentions from the message's database document? (For example if I edit from `<@a> <@b>` to `<@a>`, does `b` get removed from the document?)

docs/api.md
heyitsmeuralex: We don't have commas in [some?] other models of the API docs; plus it's not particularly standard to anyone but us. I'd like to leave it off if I'm honest.

packages/server/api.js
heyitsmeuralex: As far as I know, yes.

docs/api.md
heyitsmeuralex: Just committed to make everything more consistent.

towerofnix: The problem here is that there are more properties after this one (`authorized`, `email`).

packages/client/src/components/sidebar-right.js
```diff
@@ -175,6 +175,13 @@ const component = (state, emit) => {
               <div
                 class='Sidebar-list-item UserList-user ${user.online ? 'is-online' : 'is-offline'}'
                 title='${user.username}${user.flair ? ` {${user.flair}}` : ''} (${user.online ? 'Online' : 'Offline'})'
+                onclick=${() => {
+                  const textarea = document.querySelector('.MessageEditor-textarea')
+
+                  if (textarea) {
+                    textarea.value += '@' + user.username
```
towerofnix: This should insert a space before the `@` if (the textarea is not empty && the last character is not a space / at the start of the line).

Also, this should totally insert at wherever the cursor is positioned in the textarea, but don't worry about that now!! It's super a pain to implement.

packages/client/src/components/sidebar-right.js
```diff
@@ -179,7 +179,12 @@ const component = (state, emit) => {
                   const textarea = document.querySelector('.MessageEditor-textarea')
 
                   if (textarea) {
-                    textarea.value += '@' + user.username
+                    if (textarea.value.length > 0) {
+                      const lastChar = textarea.value[textarea.value.length - 1]
+                      if (lastChar !== ' ') textarea.value += ' '
```
towerofnix: Should be `lastChar !== ' ' && lastChar !== '\n'`.

Issue #259, opened 2018-03-02
Client is a mess


heyitsmeuralex:
**A word of warning:** typical alex below. Beware.

---

We really need to move stuff around in the client source. `emit('messages.fetch')` and being unable to detect when it finishes is horrible, etc, etc. We should use the event emitter a lot less and just use controllers to do things, eg. `require('./messages/controller').fetch(messagesEl, channel)`, or something. I don't know. It just seems super flaky at the moment and it's causing stupid bugs like how channelrefs never work and glitch out and generally go wrong. Aaaa. Not fun. 

Gonna need some advice @towerofnix.

PullJosh:
Have we considered using [Redux](https://redux.js.org/)? It can work with any view library (not just react!) and should help with organizing the barrage of state changes that need to be dealt with.

heyitsmeuralex:
Totally! Redux is cool. MobX is also cool.
Which to use? :stuck_out_tongue_winking_eye:

> We could attempt to make something similar to @TheInitializer's (currently unreleased) decent.py for Node/browsers to simplify business logic..? That could help. 

Also, I definitely want to try this.

PullJosh:
> Redux is cool. MobX is also cool. Which to use? :stuck_out_tongue_winking_eye:

I would strongly recommend going with Redux. One of the big hurdles with choo has been that nobody is familiar with it, so nobody can contribute to the client without spending a bunch of time learning how choo works.

Although the popularity difference between Redux and MobX is much less extreme than choo vs React, I still think it's something to take into consideration. I would also expect Redux to have stronger support throughout other tools/frameworks than MobX, and redux goes hand-in-hand with react *if* we choose to make the jump to Preact.

heyitsmeuralex:
Note that this is going super well in the `preact` branch. Decoupling non-UI logic into its own package was a great idea.

PR #268, opened 2018-03-02
Fix state.secure being reset to false

towerofnix:
This effectively reverts https://github.com/decent-chat/decent/commit/2c39f3cf5f731107f72300fe8f0d94a7f8f97033. Oh well.

Right now, switching between servers is kind of *really* broken; I found out while testing this. Fun.

@heyitsmeuralex Go to bed - *sleep* :tada: - after merging this!


heyitsmeuralex:
>Right now, switching between servers is kind of really broken; I found out while testing this. Fun.

Gonna say this is related to #259.

PR #277, opened 2018-03-03
decent.js

heyitsmeuralex:
Towards resolving #259. But also a thing I wanted to do anyway. See the decent.js package for more information; it's essentially what discord.js is to Discord.

Slightly more geared toward API spec 1.0.0 (eg. does not handle authorization-requiring servers as those are removed), but it works with the current 0.1.0 implementation, since there's no actual server implementing 1.0.0-preview yet.

### DIFF
--- a/packages/decent.js/docs/channel.md
+++ b/packages/decent.js/docs/channel.md
@@ -7,6 +7,7 @@ Properties:
 
 * channel.id (string)
 * channel.name (string)
+* channel.deleted (boolean)
 
 The following properties are only provided if logged in:
 
--- a/packages/decent.js/docs/channel.md
+++ b/packages/decent.js/docs/channel.md
@@ -1,0 +1,53 @@
+# const channel = client.channels.nth(0)
+Represents a single channel. Updated automatically.
+
+---
+
+Properties:
+
+* channel.id (string)
+* channel.name (string)
+
+The following properties are only provided if logged in:
+
+* channel.unreadMessageCount (?int)
+* channel.oldestUnreadMessageID (?string)
+
+---
+
+Extends [eventemitter3](https://npm.im/eventemitter3).
+
+## await channel.sendMessage(text: string) -> string messageID
+Sends a message to the channel, implicitly marking it as read.
+
+## await channel.markRead()
+Only works if logged in. Marks the channel as read for the logged-in user.
+
+## await channel.getMessages({ before: ?[Message](message.md), after: ?[Message](message.md), limit: ?int }) -> Array<[Message](message.md)>
+Resolves with an array (not set!) of [message](message.md)s of max length `limit` (default 50). Note `limit` must be `<= 50`.
+
++ If `before` is provided, the first message returned will be the message that comes right before the message `before`.
++ Similarly, if `after` is provided the first message returned will be the message directly after the message `after`.
++ If neither are provided, the latest messages are returned.
+
+## await channel.getPins() -> [PinnedMessages](pinned-messages.md)
+Resolves with a PinnedMessages set containing the messages pinned to this channel.
+
+## channel.toString()
+Returns `#channelname`.
+
+---
+
+Use `channel.on('event', callback)` method to listen for events.
+
+## event 'update' (channel: [Channel](channel.md))
+Emitted when this channel's details (`email` or `flair`) are updated.
+
+## event 'delete' (channel: [Channel](channel.md))
+Emitted when this channel is deleted.
+
+## event 'online' (channel: [Channel](channel.md))
+Emitted when this channel comes online.
+
+## event 'offline' (channel: [Channel](channel.md))
+Emitted when this channel goes offline.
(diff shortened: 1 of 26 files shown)


PullJosh:
Want a review on this or should I wait until later?

heyitsmeuralex:
A review would be cool, although it's not yet done @PullJosh. Still need to cover the rest of the API eg. deleting messages.

packages/decent.js/docs/client.md
```diff
@@ -21,6 +21,9 @@ Attempts to log in using the provided credentials. Resolves with the newly logge
 ## await client.loginWithSessionID(sessionID: string)
 Attempts to log in using the provided sessionID. Resolves with the newly logged-in [user](user.md).
 
+## await client.logout(deleteSessionID: boolean)
+Logs out. `deleteSessionID` defaults to true.
+
```
towerofnix: What is an example usage of not deleting the session ID?

heyitsmeuralex: Say if you wanted to later login using the same session ID, I guess.

towerofnix: Fair enough!

heyitsmeuralex:
This should be done now.

packages/decent.js/docs/channel.md
```diff
@@ -34,6 +34,12 @@ Resolves with an array (not set!) of [message](message.md)s of max length `limit
 ## await channel.getPins() -> [PinnedMessages](pinned-messages.md)
 Resolves with a PinnedMessages set containing the messages pinned to this channel.
 
+## await channels.rename(newName: string)
+Renames this channel.
+
+## await channels.delete()
+Deletes this channel.
+
```
towerofnix: Should these be `channel.rename` and `channel.delete` (instead of <code>channel<strong>s</code></code>)?

heyitsmeuralex: Yup.

towerofnix: Same deal with `channel.create`.

packages/decent.js/docs/channel.md
```diff
@@ -0,0 +1,60 @@
+# const channel = client.channels.nth(0)
+Represents a single channel. Updated automatically.
+
+---
+
+Properties:
+
+* channel.id (string)
+* channel.name (string)
+* channel.deleted (boolean)
```
PullJosh: When would this be used? Does the channel not just disappear when deleted?

packages/decent.js/docs/channel.md
```diff
@@ -0,0 +1,60 @@
...
+
+* channel.unreadMessageCount (?int)
+* channel.oldestUnreadMessageID (?string)
+
+---
+
+Extends [eventemitter3](https://npm.im/eventemitter3).
+
+## await channel.sendMessage(text: string) -> string messageID
+Sends a message to the channel, implicitly marking it as read.
```
PullJosh: Could it be valuable to provide a boolean setting to enable/disable the implicit marking as read? Obviously defaulting to true would be best, but there's no reason not to give clients the option.

heyitsmeuralex: >no reason

The _server_ marks it as read, not us.
It's not an API option.

packages/decent.js/docs/channel.md
heyitsmeuralex: If you have already have a Channel and it gets deleted, it is removed from `client.channels` but your reference can't just become null; `channel.deleted = true` is done instead.

```js
const channel = client.channels.find(c => c.name === 'general')

await channel.delete()

channel.deleted //=> true
client.channels.find(c => c.name === 'general') //=> undefined
```

packages/decent.js/docs/channel.md
PullJosh: Should it be?

packages/decent.js/docs/channel.md
PullJosh: Ah. I suppose that makes sense.

packages/decent.js/docs/channel.md
heyitsmeuralex: @towerofnix

packages/decent.js/docs/client.md
```diff
@@ -0,0 +1,62 @@
+# const Client = require('decent.js')
+Constructor takes no arguments:
+
+```js
+const Client = require('decent.js')
```
PullJosh: I feel like it would easier to read if this line were:
```js
const Decent = require('decent.js')
```

Having two variables of the same name but with different capitalization is bound to cause confusion.

heyitsmeuralex: That's a really common pattern :thinking:

PullJosh: Huh. Never seen it. :package:

packages/decent.js/docs/channel.md
towerofnix: Except for bots, what is the practical reason of *not* marking a channel as read once you send a message? @PullJosh

heyitsmeuralex:
Still need an _actual review_ (merge, ask for changes) of this please, guys.

PullJosh:
I'm sure @towerofnix would be *thrilled* to do it. :stuck_out_tongue:

PullJosh:
Login/out event please!

heyitsmeuralex:
@PullJosh done in `preact` branch.

heyitsmeuralex:
Hi can this get merged please

PR #285, opened 2018-03-03
Fix XSS for Decent-flavoured links in Markdown

joker314:
This PR is ***untested***; and I'm afraid I won't be able to get a local copy up due to memory restrictions (I might try to do so at some point, but I'd prefer someone else with more disk space do so)

**Note**: *Please observe that this PR is for the `xss` branch, not the `master`. Is that a bad idea?*


joker314:
I notice no changes on meta.decent.chat and can still reproduce all the flaws

Issue #281, opened 2018-03-03
Implement API spec 1.0.0


heyitsmeuralex:
[The spec](https://github.com/decent-chat/decent/blob/1.0.0/docs/api.md).

heyitsmeuralex:
@towerofnix can we get an ETA on this? This issue is pretty blocking of most other server-related issues (eg. #286)

towerofnix:
@heyitsmeuralex ETA is "eventually". I'm busy with school.

But it'll definitely go faster if we have a checklist on what needs to be done - the API spec was updated, but not in too many colossal ways. I think the biggest change was just roles and permissions. What other things need to be done?

heyitsmeuralex:
>life and school

eek

---

Aside from those PRs above (thanks!):

* roles/permissions
* properties are now part of /api/
* settings event

towerofnix:
Getting rid of "decent.js" here since it's not relevant - that's included in preact, not 1.0.0. This issue is just for tracking progress on the server (basically only roles/permissions now, AFAIK).

PR #295, opened 2018-03-08
Implement Unix timestamps

towerofnix:
Towards #281. This makes dates generally be stored as Unix timestamps, as per the 1.0.0 specification. It also updates the client to deal with this. (Conveniently, changing `second = 1000` to `second = 1` did half the work automatically!)


heyitsmeuralex:
Merged into `1.0.0`.

PR #296, opened 2018-03-08
Add basic type property to messages

towerofnix:
Towards #281. All messages have a type property that is `"user"`, for now. System-messages can be implemented later and are not a blocker for continuing general work. It'll be best to do them after permissions are implemented.


heyitsmeuralex:
LGTM, but going to merge this into the `1.0.0` branch rather than master.

PR #299, opened 2018-03-08
1.0.0 impl: various datetime fixes

### LINKED ISSUE #297: Should message dates *really* not be to the millsecond?
Right now some tests are failing because the order of messages sent in the same second isn't obvious to the server - as far as it's concerned, those messages were sent within the same instant. And they *were* sent in the same unit of time - since our unit of time is whole seconds.

Would it be okay to simply remove the `Math.round`ing of `(javascript date) / 1000`? That would keep the accuracy to the millisecond, but still use seconds as the unit equal to 1.

Ping @heyitsmeuralex for thoughts; this is an extremely easy fix.

heyitsmeuralex:
* Updated message datetime keys as per spec (more consistent with the rest of the API):
  * `message.date` -> `message.dateCreated`
  * `message.editDate` -> `message.dateEdited`
  * Updates tests to reflect
* Fix #297 
  - We now store dates to millisecond-level accuracy, however the serialize functions floor the value before they are outputted by the API.
    + This fixes some tests.
  - We may want to get the spec to specify whether these dates should always be integers (what we're doing now) or may be floats.
  - Note that currently the API specifies "number" as the type. This is ambiguous, and we should probably use 'integer' and 'float' across the entire spec if applicable.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -334,7 +334,7 @@ DELETE /api/emotes/package
 ```js
 {
   "id": string,
-  "dateCreated": number // Unix time at creation
+  "dateCreated": float // Unix time at creation. May be more accurate using decimal places
 }
 ```
 
@@ -436,9 +436,9 @@ DELETE /api/sessions/12345678-ABCDEFGH
   "authorAvatarURL": string,
 
   // Dates are returned as the number of seconds since UTC 1970-1-1, commonly
-  // known as Unix time.
-  "dateCreated": number,
-  "dateEdited": number | null,
+  // known as Unix time. May be more accurate using decimal places.
+  "dateCreated": float,
+  "dateEdited": float | null,
 
   "pinned": boolean,
   "mentionedUserIDs": [ ID ]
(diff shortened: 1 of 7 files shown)


towerofnix:
>We now store dates to millisecond-level accuracy, however the serialize functions floor the value before they are outputted by the API.

Why?

We're a *chat service*. This means people send messages *quickly* - very often multiple times in a single second. Why should we prevent the user (or developers) from knowing the order of messages that came within a single second?

I know that some languages have int types which are separate from floats, but they can just convert the values to integers, if they really need to.

heyitsmeuralex:
>Why should we prevent the user (or developers) from knowing the order of messages that came within a single second?

Unix time is standard. `Date.now()` is not. The milliseconds float thing should be specified in the API as a recommended (but not required) practice imo.

PullJosh:
> How many programming languages do not give a way to tell the current millisecond?

When we write a server implementation in [BF](https://en.wikipedia.org/wiki/Brainfuck), we'll have a bit of thinking to do...

PR #301, opened 2018-03-08
Improve 1.0.0-spec date accuracy

### LINKED ISSUE #297: Should message dates *really* not be to the millsecond?

towerofnix:
Resolves #297.


towerofnix:
Oops, looks like you've already made a PR! *looks*

